PR #5, opened 2018-07-27
Open App settings in iOS

mvanbeusekom:
Added the `openAppSettings` API to iOS (note this PR adds to #4)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,38 @@
+matrix:
+  include:
+    - os: linux
+      env:
+        - SHARD="Build example apks"
+      jdk: oraclejdk8
+      sudo: false
+      addons:
+        apt:
+          # Flutter depends on /usr/lib/x86_64-linux-gnu/libstdc++.so.6 version GLIBCXX_3.4.18
+          sources:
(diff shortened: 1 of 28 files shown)


dark-chocolate:
`bool canOpenSettingsPage = await PermissionHandler().openAppSettings();
`
Doc says, it returns `true` if settings page can be opened and `false` otherwise. However in iOS, it opens settings page but still `false` is returned. So, please correct your docs.

mvanbeusekom:
@dark-chocolate, first the docs don't mention anything specific regarding the boolean being returned (which is definitely something we could/ need to approve on, pull-request are very welcome).

Second the idea of the boolean is not to indicate if the settings "can" be opened, but if it "was" opened successfully, meaning if `true` is returned the page has been opened and if `false` is returned the page could not be opened. Since you mentioned on iOS the a `false` is returned but the page is still opened I would say this is a bug and we need to look into that. To keep track of this bug I have create issue #52. Please feel free to add findings or other details there.

dark-chocolate:
@mvanbeusekom  Is there also a way to open Settings page in iOS (Settings here doesn't mean the my own app settings, it means the general Settings app of iPhone)?

PR #18, opened 2018-09-03
Make methods non static

martijn00:

Feature. You can now override or make a new instance of the plugin

All methods are static

Not anymore

Yes

- [ ] All projects build
- [ ] Follows style guide lines ([code style guide](https://github.com/BaseflowIT/flutter-geolocator/blob/develop/CONTRIBUTING.md))
- [ ] Relevant documentation was updated
- [ ] Rebased onto current develop

### DIFF
--- a/lib/permission_handler.dart
+++ b/lib/permission_handler.dart
@@ -14,7 +14,7 @@ part 'package:permission_handler/utils/codec.dart';
 class PermissionHandler {
   factory PermissionHandler() {
     if (_instance == null) {
-      final MethodChannel methodChannel =
+      const MethodChannel methodChannel =
           const MethodChannel('flutter.baseflow.com/permissions/methods');
 
       _instance = new PermissionHandler.private(methodChannel);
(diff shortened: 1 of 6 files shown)


lib/permission_handler.dart
```diff
@@ -5,19 +5,34 @@ import 'dart:convert';
...
 part 'package:permission_handler/permission_enums.dart';
 part 'package:permission_handler/utils/codec.dart';
 
 /// Provides a cross-platform (iOS, Android) API to request and check permissions.
 class PermissionHandler {
-  static const MethodChannel _channel =
-      const MethodChannel('flutter.baseflow.com/permissions/methods');
+  factory PermissionHandler() {
+    if (_instance == null) {
+      final MethodChannel methodChannel =
```
mvanbeusekom: @martijn00, the static code analyses is complaining about this line and mentions it is better to use `const` over `final` (which is correct):

"info • Prefer const over final for declarations • lib/permission_handler.dart:17:7 • prefer_const_declarations"

So could you change this to:

```
const MethodChannel methodChannel =
    const MethodChannel('flutter.baseflow.com/permissions/methods');
```

PR #62, opened 2019-01-22
 Make sure to return permission state over service state

### LINKED ISSUE #54: iOS gives more preference to disabled status then denied status
## 🐛 Bug Report
If location services is off, and your app has not given permission, and you check for permission status, then disabled status is returned instead of denied. 

### Expected behavior
The correct behavior would be it should first check if the permission is granted and if the answer is yes, then it should check for location services if they are on or off. 

### Reproduction steps
Make sure the location services is off (GPS) in iOS and your app has not asked for permission before. Now ask for the permission very first time, the status returned would be `PermissionStatus.disabled` instead of `PermissionStatus.denied` or `PermissionStatus.unknown`. However if location services is on and you ask for permission then it works as expected and returns `PermissionStatus.unknown`

`PermissionStatus status = await PermissionHandler().checkPermissionStatus(PermissionGroup.locationAlways);
`

### Result

status = `PermissionStatus.disabled` (if location is off not matter if permission is given to your app or not) -- NOT EXPECTED, the status should first check if permission is granted or not, if permission is granted then it should check if location is disabled or not.  

status = `PermissionStatus.unknown` (if location is on and permission is not asked before) -- EXPECTED

**Version:** 2.1.2

**Platform:** 
iOS 12.0.1 running in iPhone 6s

mvanbeusekom:

Bug fix

At the moment the services status has precedence over the permission status. Meaning when the service is down you don't know if the user has granted permissions to access the services.

The permissions status has precedence over the service status. Meaning the plugin will first check if the user has given permission and only if so (`PermissionStatus.granted`), the plugin will check if the service is available (if the service is not available the plugin will return a `PermissionStatus.disabled` status, meaning the user has granted access but the service is not running).

No

Run the example app

- Fixes #54 

- [x] All projects build

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+## 2.1.1
+
+* Update to the latest version of Swift (4.2);
+* Make sure that the correct Swift version is set in the `Podfile` of consuming Apps;
+* Updated configuration for statis code analyses, so it complies with the Flutter recommendations.
+
 ## 2.1.0
 
 * Added Android support to check if location services are enabled. If location services are not running the permission check returns `PermissionStatus.DISABLED`.
--- a/README.md
+++ b/README.md
@@ -22,7 +22,7 @@ To use this plugin, add `permission_handler` as a [dependency in your pubspec.ya
 
 ```yaml
 dependencies:
-  permission_handler: '^2.1.0'
+  permission_handler: '^2.1.1'
 ```
 
 > **NOTE:** There's a known issue with integrating plugins that use Swift into a Flutter project created with the Objective-C template. See issue [Flutter#16049](https://github.com/flutter/flutter/issues/16049) for help on integration.
--- a/README.md
+++ b/README.md
@@ -143,7 +143,7 @@ enum PermissionStatus {
   /// Permission to access the requested feature is denied by the user.
   denied,
 
-  /// The feature is disabled (or not available) on the device.
+  /// Permissions to access the feature is granted by the user but the feature is disabled.
   disabled,
 
   /// Permission to access the requested feature is granted by the user.
--- a/analysis_options.yaml
+++ b/analysis_options.yaml
@@ -28,6 +28,10 @@ analyzer:
     missing_return: warning
     # allow having TODOs in the code
     todo: ignore
+    # Ignore analyzer hints for updating pubspecs when using Future or
+    # Stream and not importing dart:async
+    # Please see https://github.com/flutter/flutter/pull/24528 for details.
+    #sdk_version_async_exported_from_core: ignore
   exclude:
     - 'bin/cache/**'
     # the following two are relative to the stocks example and the flutter package respectively
@@ -56,6 +60,7 @@ linter:
     - avoid_empty_else
     - avoid_field_initializers_in_const_classes
     - avoid_function_literals_in_foreach_calls
+    # - avoid_implementing_value_types # not yet tested
     - avoid_init_to_null
     # - avoid_js_rounded_ints # only useful when targeting JS runtime
     - avoid_null_checks_in_equality_operators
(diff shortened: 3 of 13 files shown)


dark-chocolate:
@mvanbeusekom Thanks for your great work, it is working as expected. However I do have one problem now. Please consider the following case. 

Let's say, user goes to my App settings (in iOS settings page) and denies the location permission.  (For this case, when I say location I mean `LocationWhenInUse`) So, when he comes back to my app, I check for permission and the permission status = `PermissionStatus.denied` so far so good. But if the permission is denied, how do I know if location service is ON/OFF, if location service is OFF, I can ask user to turn on location services, because without turning on location services, user can't grant the location permission (weird but that's how iOS works and which is why I love Android - simple and powerful :)

Simply put, I want to say, your `geolocator` plugin needs to work here, by checking 
```
Geolocator().checkGeolocationPermissionStatus(locationPermission: GeolocationPermission.locationWhenInUse)
```

 However this plugin hasn't been updated. So, you guys need to sync that plugin with the current `permission_handler` plugin to make both plugins cover every case.  

Thanks again for fixing bugs quickly 💯

mvanbeusekom:
@dark-chocolate thank you very much for you feedback (and patience). I see your problem and am thinking of a solution, maybe you can help verify if it helps. 

In iOS the following method is handling your call to `checkGeolocationPermissionStatus`:

```swift
    private static func getPermissionStatus(permission: PermissionGroup) -> PermissionStatus {
        let authorizationStatus: CLAuthorizationStatus = CLLocationManager.authorizationStatus()
        let permissionStatus: PermissionStatus = LocationPermissionStrategy.determinePermissionStatus(
            permission: permission,
            authorizationStatus: authorizationStatus)
        
        if permissionStatus == PermissionStatus.granted && !CLLocationManager.locationServicesEnabled() {
            return PermissionStatus.disabled
        }
        
        return permissionStatus
    }
```

As you can see the code only returns `PermissionStatus.disabled` when the location permissions are granted and the location services are disabled. I am thinking to update the if-statement to:

```swift
        if (permissionStatus == PermissionStatus.granted || permissionStatus == PermissionStatus.denied) && !CLLocationManager.locationServicesEnabled() {
            return PermissionStatus.disabled
        }
```

Meaning the `disabled` state is returned when the location permissions are either `granted` or `denied`  and the location services are disabled. 

This allows you to still ask for permissions when the services are disabled (if permissions have not been set yet, so `PermissionStatus.unknown`). 

Next to that I will also include an extra method `isServiceEnabled(PermissionGroup permission)` which will return one of the following values:

```
ServiceStatus.unkown // Used when there is no service for the supplied permission group or state could not be determined
ServiceStatus.enabled // The services for the permission group are enabled
ServiceStatus.disables // The services for the permission group are disabled
``` 

Any feedback would be greatly appreciated, and depending on your feedback I will try to make an implementation this week.

dark-chocolate:
@mvanbeusekom Thanks for your prompt response. Actually I did give it a try (`if` condition you mentioned) but it didn't work. 

**Reason**: As I mentioned earlier, the current behaviour is returning `GeolocationStatus.denied` but expected behaviour should return `GeolocationStatus.disabled` so the code you wrote isn't relevant to make changes (as far as I can tell). 

We need to replace `denied` with `disabled` but where that I don't know because I am not familiar with iOS and don't understand any sort of syntax there. So, I just gave you a hint where code needs to be changed, however I am happy to test any kind of code you provide me. 

Thanks again!

mvanbeusekom:
@dark-chocolate, no problem I will make my suggested changes in Swift and test it locally first. I should be able to provide an update in a few hours.

mvanbeusekom:
@dark-chocolate we just release version 2.2.0 of this plugin which should solve you problems. If the location services are disabled you will get a `PermissionStatus.disabled` response when checking permissions (no matter if currently permissions are denied or granted). 

I also included a new method `checkServiceStatus(PermissionGroup permission)` which can be used to get the current state of the location services or motion services (depending on the supplied value). See also the README.md for more details.

dark-chocolate:
@mvanbeusekom  Thanks a lot buddy, didn't know you guys are this quick in fixing issues. 🥇 💯

PR #69, opened 2019-02-08
specify the swift version

ganisback:
Specify the swift version to avoid compile error when use cocoapods 1.6


ganisback:
below is pod install error if not specify swift version:
```
- `permission_handler` does not specify a Swift version and none of the targets (`Runner`) integrating it have the `SWIFT_VERSION`
    attribute set. Please contact the author or set the `SWIFT_VERSION` attribute in at least one of the targets that integrate this pod.
```

Issue #111, opened 2019-05-16
Android PERMISSION_CODE is the same as location_permissions PERMISSION CODE


ened:
## 🔙 Regression

This will cause conflicts with the latest geolocator plugin because it uses the location_permissions package and thus we can not retrieve location permission request results in here when both projects are used.

Will submit a new PR which uses a different request code:

 - https://github.com/BaseflowIT/flutter-permission-plugins/blob/develop/packages/location_permissions/android/src/main/java/com/baseflow/location_permissions/LocationPermissionsPlugin.java#L38
 - https://github.com/BaseflowIT/flutter-permission-handler/blob/develop/android/src/main/java/com/baseflow/permissionhandler/PermissionHandlerPlugin.java#L36

### Configuration

**Version:** 3.0.2

**Platform:** 
- [ ] :iphone: iOS
- [x] :robot: Android

mvanbeusekom:
I will wait for this PR before rolling a new release.

Are we free to pick any request code we want? I choose 25 because I read it somewhere in the Google Documentation and to be honest I didn't really investigate the use of the code, figured it was some kind of Android constant that had to be used.

ened:
Fixed in #113.

PR #113, opened 2019-05-16
Android uses request code 24 from now on

ened:

To prevent issues with the location_permissions plugin, as it's being used from geolocator.

Conflicts with location_permissions.. details in #111 

Change request code.

include geolcoator in example - test permissions like call or storage,

Consider to set proper request codes for all BaseflowIT plugins to avoid any further conflicts.

- [x] All projects build

### DIFF
--- a/android/src/main/java/com/baseflow/permissionhandler/PermissionHandlerPlugin.java
+++ b/android/src/main/java/com/baseflow/permissionhandler/PermissionHandlerPlugin.java
@@ -33,7 +33,7 @@ import io.flutter.plugin.common.PluginRegistry.Registrar;
 
 public class PermissionHandlerPlugin implements MethodCallHandler {
   private static final String LOG_TAG = "permissions_handler";
-  private static final int PERMISSION_CODE = 25;
+  private static final int PERMISSION_CODE = 24;
 
   //PERMISSION_GROUP
   private static final int PERMISSION_GROUP_CALENDAR = 0;


ened:
@mvanbeusekom consider cutting a public release soon and also amend the README about the plugin conflict

mvanbeusekom:
@ened really grateful for your pull-request, however could you elaborate on your remark:

> Consider to set proper request codes for all BaseflowIT plugins to avoid any further conflicts.

What do you mean with "proper"? Is there any additional documentation I can read through to make sure this doesn't happen again and also not conflicts with other plugins (not by Baseflow)?

ened:
You can not avoid conflicts with other plugins. Root cause is this line here:

https://github.com/BaseflowIT/flutter-permission-handler/blob/develop/android/src/main/java/com/baseflow/permissionhandler/PermissionHandlerPlugin.java#L122

Both the `location_permissions` plugin and `permission_handler` register the same request code. However, `location_permissions` get's called earlier and thus eats the permission request result (by returning true, so Flutter won't deliver the event to the next plugin that is interested in the permission result), in this case `permission_handler`.

By proper I meant you should introduce a 3rd party dependency to hold all of the BaseflowIT constants (all the request codes).

I also would not worry about other plugins. How many permission handling plugins can a developer install & use? ;-).
The numbers themselves should be a bit higher I think.

ened:
BTW, the bit about 3rd party library to keep all the constants assumes that you probably want to start  a collection of plugins to handle each permission individually. 

permission_handler currently contains code for every permission and most notably, requires iOS developers to put permission strings in their configuration even when they don't need certain permissions. So splitting it up would be very welcome.

mvanbeusekom:
Got it, thank you for the explanation. 

The idea is indeed to phase out / replace the `permission_handler` by individual packages that will be available in the `baseflowit/flutter-permission-plugins` repository. I will definitely consider including a separate library containing request codes for each individual package their.

ened:
@mvanbeusekom just noticed I have opened PRs against the master branch.. guess you could merge develop into master once and I will open future PRs against develop, if needed. (hopefully not, though!)

ened:
@mvanbeusekom and #108 made it to develop. Sorry for that mess.

mvanbeusekom:
No worries I will fix it on my end ;)

PR #114, opened 2019-05-17
Phone service status on iOS & Android

ened:

Feature

checkServiceStatus for phone call capability will always return `notApplicable`, independent of the users' handsets capability to place phone calls.

service status of `enabled` when the handset could in theory place calls.
service status of `notApplicable` when the handset has no ability to call.

This should be (hence the WIP) more fine grained and also check if there's a actual SIM card inserted.

No, as potentially nobody is using getServiceStatus for phone calls

Use a wifi-only tablet plus 2 test phones to explore the different results.

TBD Android: https://stackoverflow.com/questions/5196833/android-device-phone-call-ability
TBD iOS:
https://stackoverflow.com/questions/5094928/detecting-whether-or-not-device-support-phone-calls/11595365#11595365

@mvanbeusekom Need your comment please if this functionality is welcome in the library.

- [x] All projects build

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## next
+
+* Support service status inquiry for phone permission on iOS & Android.
+
 ## 3.0.2
 
 * Fixed bug when rapidly requesting permissions (#23);
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -9,7 +9,7 @@ buildscript {
     }
 
     dependencies {
-        classpath 'com.android.tools.build:gradle:3.4.0'
+        classpath 'com.android.tools.build:gradle:3.4.1'
     }
 }
 
@@ -35,8 +35,8 @@ android {
 }
 
 dependencies {
-    implementation 'androidx.annotation:annotation:1.0.1'
-    implementation 'androidx.core:core:1.0.1'
+    implementation 'androidx.annotation:annotation:1.0.2'
+    implementation 'androidx.core:core:1.0.2'
 }
 
 repositories {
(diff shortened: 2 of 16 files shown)


mvanbeusekom:
@ened, this functionality is definitely welcome in this library and we will include it also in the new permission plugins when appropriate.

ened:
Will work on iOS next week.. we also need functionality there.

@mvanbeusekom Our App currently also needs to check for mobile connectivity service status (e.g. if mobile data is turned on). I suppose this could go into a separate PR.

mvanbeusekom:
@ened  regarding the connectivity there is already a great plugin from Flutter themselves, which is very feature rich. I don't really feel this belongs here with the permissions. Or am I missing something?

ened:
@mvanbeusekom The `connectivity` pub only supports the check of internet connectivity. I would like permission_handler to contain a service check for a particular Android setting, as shown here: https://stackoverflow.com/a/29625597/375209

Basically, the code required is this:
		boolean mobileDataAllowed = Settings.Secure.getInt(mActivity.getContentResolver(), "mobile_data", 1) == 1;

mvanbeusekom:
@ened, I see what you mean and in general I am not against it. But how would you request this, does that mean we need to introduce a new permission group which will only work with the `checkService` method?

Anyway let's move it indeed to a new pull-request so we can move on with this one. Would you like me to merge this PR now and you'll make a seperate one for the iOS part?

ened:
Please do not merge yet, iOS port is coming some time this week.

ened:
One more commit then ready to merge.

ened:
> @ened, I see what you mean and in general I am not against it. But how would you request this, does that mean we need to introduce a new permission group which will only work with the `checkService` method?

For mobile_data, it would  be a permissionGroup with serviceStatus inquiry only, similar to how this PR implements it on the iOS side. You can not request call permission on iOS, but you can, to some extend, inquire about the ability to place calls (*service status*).

ened:
@mvanbeusekom ping

mvanbeusekom:
@ened, sorry last two days were a bit busy. I will review this PR this afternoon.

mvanbeusekom:
@ened, I have just released version 3.1.0 which includes this change. Thanks again.

PR #116, opened 2019-05-21
New permission for ignoring battery optimizations on Android

ened:

Feature to add a new permission called "ignore battery optimizations" on Android.
Details can be found here: https://developer.android.com/training/monitoring-device-state/doze-standby#support_for_other_use_cases

Currently the `permission_handler` plugin does not support this permission.

Screenshot_20190521-154658

It's kind of a weird permission.
Just like regular permissions:
 - It is required to have a entry in the manifest to show the dialog
 - Users need to opt in in the prompt
 - Users can allow or deny the request
 - Users can revoke it at any time 

Unlike regular permissions:
 - The response to the request is delivered in a regular intent vs. a permission listener callback
 - Existing permission infrastructure (e.g. the methods in the support library) don't apply

Added a new permission named `ignoreBatteryOptimizations`.

No, but it is based on #114 which should be merged first.

Get a Android phone running Android 6 or above and test the permission in the example App.

https://developer.android.com/training/monitoring-device-state/doze-standby#support_for_other_use_cases

I think we should put links to the relevant documentation (https://developer.android.com/training/monitoring-device-state/doze-standby#whitelisting-cases) as well.

### WIP

- [ ] Document the feature publicly in the Dart API. Ensure to mention that this is subject to Play Store removal if the App does not quality for the permissions.
- [ ] This permission is currently not playing well with others when requested together. That's because the result of the request is delivered in different API methods currently. The Java plugins final `result` output should probably be centralized.

- [x] All projects build
- [ ] Relevant documentation was updated


android/src/main/java/com/baseflow/permissionhandler/PermissionHandlerPlugin.java
```diff
@@ -249,10 +268,23 @@ private int checkPermissionStatus(@PermissionGroup int permission) {
...
 
-    final boolean targetsMOrHigher = context.getApplicationInfo().targetSdkVersion >= android.os.Build.VERSION_CODES.M;
+    final boolean targetsMOrHigher = context.getApplicationInfo().targetSdkVersion >= VERSION_CODES.M;
 
     for (String name : names) {
       if (targetsMOrHigher) {
+        if (permission == PERMISSION_GROUP_IGNORE_BATTERY_OPTIMIZATIONS) {
+          String packageName = context.getPackageName();
+          PowerManager pm = (PowerManager) context.getSystemService(Context.POWER_SERVICE);
+          if (VERSION.SDK_INT >= VERSION_CODES.M) {
```
mvanbeusekom: This `if` statement seems redundant since there is already a check which checks if the current Android version is M or higher on line 274.

android/src/main/java/com/baseflow/permissionhandler/PermissionHandlerPlugin.java
```diff
@@ -565,6 +621,12 @@ private boolean openAppSettings() {
           permissionNames.add(Manifest.permission.WRITE_EXTERNAL_STORAGE);
         break;
 
+      case PERMISSION_GROUP_IGNORE_BATTERY_OPTIMIZATIONS:
+        if (hasPermissionInManifest(Manifest.permission.REQUEST_IGNORE_BATTERY_OPTIMIZATIONS))
+          if (VERSION.SDK_INT >= VERSION_CODES.M)
```
mvanbeusekom: Minor optimization here might be to switch the two if statements around. It seems to me that checking the permission manifest would be a heavier operation then checking the Android version. Maybe this would already suffice:

```
if (VERSION.SDK_INT >= VERSION_CODES.M && hasPermissionInManifest(Manifest.permission.REQUEST_IGNORE_BATTERY_OPTIMIZATIONS)) {
  permissionNames.add(Manifest.permission.REQUEST_IGNORE_BATTERY_OPTIMIZATIONS);
}
```

android/src/main/java/com/baseflow/permissionhandler/PermissionHandlerPlugin.java
ened: The first check only covers the Apps AndroidManifest.xml. The second check (Version.SDK_INT) is a runtime check for the actual Android version currently running. Therefore, I'd suggest to comment & stick with the way.

WIP: not anymore for our side, as it's in production apps.

android/src/main/java/com/baseflow/permissionhandler/PermissionHandlerPlugin.java
ened: Yes, that will be easier.

Issue #96, opened 2019-04-11
Android's "don't ask again"


hugocbpassos:
## 🚀 Feature Requests

Android's permission dialog gives the user the option to "Don't Ask Again". 

I must know, through  the `PermissionStatus` enum, if the user selected that option. That's because, if I can't ask again, I must tell the user to go to the device's settings screen to give me the necessary permissions.

I suggest creating `PermissionStatus.dontAskAgain`.

celiyawong:
Ibid.

hugocbpassos:
@celiyawong That's exactly my point.

mvanbeusekom:
@hugocbpassos, @celiyawong, you can use the `shouldShowRequestPermissionRationale` method to figure this out. 

This method returns true if the user has previously denied the request. But if the user has selected “Dont’ ask again” this method would always return false. It also returns false if we are prompting for permission for the first time.

hugocbpassos:
@mvanbeusekom So the only way I can be 100% sure that the user selected "Don't ask again" is **requesting** the permission, checking if it's denied and then checking if `shouldShowRequestPermissionRationale` returns `false`, right? If I just want to **check** the permission, the same logic wouldn't work because if I haven't **requested** the permission at least once, it'll be denied and `shouldShowRequestPermissionRationale` would return `false` either. I would have to rely on my own code to know if the permission had been requested before or stop **checking** permissions, but always **requesting**.

I'm sorry, but you got to agree with me that this is a bad approach. Not only because it wouldn't work in the mentioned situation, but also because it makes things more complex and demands the plugin user to create more platform-specific code – `if (Platform.isAndroid) doThis() else doThat()`.

I would be glad if you considered adding the `dontAskAgain` to `PermissionStatus` enum.

Please, feel free to correct me if something I said is wrong.

mvanbeusekom:
So to find out if the user selected “don’t ask again” (even the first time you are requesting permissions) is (this is on Android):

1. Make a call to the `checkPermissions`
2. If the return value is `denied` call the `shouldShowPermissionRationale`:
a. If the value returned equals `false`, the user selected "don’t ask again".
b. If the value equals `true` it is either the first time you are requesting permissions or the user denied permissions earlier but did NOT select the “don’t ask again” option. Either way it is save to continue requesting permissions.

So you don’t have to “always” request permissions.

hugocbpassos:
[...]

Issue #128, opened 2019-06-18
shouldShowRequestPermissionRationale always false on Android


altherat:
## 🐛 Bug Report

`shouldShowRequestPermissionRationale` is always returning `false` on Android. I am trying to check for storage permissions and I've put READ_EXTERNAL_STORAGE and WRITE_EXTERNAL_STORAGE permissions in my AndroidManifest. Nothing is logged in the debug console and it happens on both debug and release versions.

### Expected behavior

`shouldShowRequestPermissionRationale` to return `true` if permissions have not been accepted on Android. Or am I misunderstanding something?

### Reproduction steps

Sample app:
```
import 'package:flutter/material.dart';
import 'package:permission_handler/permission_handler.dart';

void main() => runApp(App());

class App extends StatelessWidget {
  
  @override
  Widget build(BuildContext context) {
    return MaterialApp(
      home: Scaffold(
        floatingActionButton: FloatingActionButton(
          onPressed: () async {
            final PermissionHandler permissionHandler = PermissionHandler();
            final bool show = await permissionHandler.shouldShowRequestPermissionRationale(PermissionGroup.storage);
            print(show);
            final PermissionStatus permissionStatus = await permissionHandler.checkPermissionStatus(PermissionGroup.storage);
            print(permissionStatus);
          }
        ),
      )
    );
  }

}
```
output:
```
false
PermissionStatus.denied
```

### Configuration

**Version:** 3.1.0
**Flutter:** 1.5.4-hotfix.2
**Dart:** 2.3.0
**Device:** OnePlus 6t, Android Pie

**Platform:** 
- [ ] :iphone: iOS
- [x] :robot: Android

**Edit:** It appears to return `true` after I've used `requestPermissions` and pressed the deny button on the dialog. But it returns `false` again after I clear the app's storage. For now I can workaround by just using `checkPermissionStatus` to see if it is `PermissionStatus.denied`, but I'm not sure how it's handled on iOS.

mvanbeusekom:
In the next version (4.2.0) we will introduce a fix for this. When using `checkPermissionStatus` or `requestPermissions` you will (only on Android) receive an extra result item "neverAskAgain" in those cases were the user denied permissions and checked the "Don't ask again" checkbox.

PR #189, opened 2020-01-12
Added neverAskAgain permission status for android devices

ManoyloK:
Added new permission state on android devices when user denied permission request with selected checkbox "Never ask again"

When user "Never ask again" it returns  status of  permission "denied", so we can't understand actual state of permission and after permission request we got nothing 

When user "Never ask again" it returns new permission status neverAskAgain

https://github.com/Baseflow/flutter-permission-handler/issues/96

- [x] All projects build

### DIFF
--- a/android/src/main/java/com/baseflow/permissionhandler/PermissionHandlerPlugin.java
+++ b/android/src/main/java/com/baseflow/permissionhandler/PermissionHandlerPlugin.java
@@ -19,6 +19,7 @@ import android.telephony.TelephonyManager;
 import android.text.TextUtils;
 import android.util.Log;
 
+import androidx.annotation.RequiresApi;
 import androidx.core.app.NotificationManagerCompat;
 import io.flutter.plugin.common.PluginRegistry.ActivityResultListener;
 import java.lang.annotation.Retention;
@@ -100,7 +101,7 @@ public class PermissionHandlerPlugin implements MethodCallHandler {
   private static final int PERMISSION_STATUS_GRANTED = 2;
   private static final int PERMISSION_STATUS_RESTRICTED = 3;
   private static final int PERMISSION_STATUS_UNKNOWN = 4;
-
+  private static final int PERMISSION_STATUS_NEWER_ASK_AGAIN = 5;
   @Retention(RetentionPolicy.SOURCE)
   @IntDef({
       PERMISSION_STATUS_DENIED,
@@ -108,6 +109,7 @@ public class PermissionHandlerPlugin implements MethodCallHandler {
       PERMISSION_STATUS_GRANTED,
       PERMISSION_STATUS_RESTRICTED,
       PERMISSION_STATUS_UNKNOWN,
+      PERMISSION_STATUS_NEWER_ASK_AGAIN,
   })
   private @interface PermissionStatus {
   }
@@ -204,12 +206,6 @@ public class PermissionHandlerPlugin implements MethodCallHandler {
     }
   }
 
-  private final Registrar mRegistrar;
-  private Result mResult;
-  private ArrayList<String> mRequestedPermissions;
-  @SuppressLint("UseSparseArrays")
-  private Map<Integer, Integer> mRequestResults = new HashMap<>();
-
   @Override
   public void onMethodCall(MethodCall call, Result result) {
     switch (call.method) {
@@ -302,7 +298,9 @@ public class PermissionHandlerPlugin implements MethodCallHandler {
         }
         final int permissionStatus = ContextCompat.checkSelfPermission(context, name);
         if (permissionStatus == PackageManager.PERMISSION_DENIED) {
-          return PERMISSION_STATUS_DENIED;
+          if (VERSION.SDK_INT >= VERSION_CODES.M && isNeverAskAgainSelected(permission)) {
+            return PERMISSION_STATUS_NEWER_ASK_AGAIN;
+          } else return PERMISSION_STATUS_DENIED;
         } else if (permissionStatus != PackageManager.PERMISSION_GRANTED) {
           return PERMISSION_STATUS_UNKNOWN;
         }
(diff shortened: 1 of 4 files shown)


ManoyloK:
Hi @mvanbeusekom shouldShowRequestPermissionRationale() returns true if the user has previously denied the request, and returns false if a user has denied a permission and selected the Don't ask again and it also return false when application launches at first time. So we have situation when permission is denied and shouldShowRequestPermissionRationale returns false when permission has not been requested yet and when user has denied a permission and selected the Don't ask again. Due to the fact that we can't understand is system will  show permission dialog or no

ManoyloK:
> Hi @ManoyloK, can you explain why we need the extra value `neverAskAgain` and how it differs from the `shouldShowRequestPermissionRationale()` method?
> 
> It is my understanding that you can use this method to ask Android if the user has denied the permissions and selected the **Don't ask again** option (see also [here](https://developer.android.com/training/permissions/requesting#explain) and [here](https://developer.android.com/reference/androidx/core/app/ActivityCompat.html#shouldShowRequestPermissionRationale(android.app.Activity,%20java.lang.String)))
> 
> It could be I am missing something so please feel free to correct me.

Hi @mvanbeusekom ```shouldShowRequestPermissionRationale()``` returns true if the user has previously denied the request, and returns false if a user has denied a permission and selected the Don't ask again and it also return false when application launches at first time. So we have situation when permission is denied and shouldShowRequestPermissionRationale returns false when permission has not been requested yet and when user has denied a permission and selected the Don't ask again. Due to the fact that we can't understand is system will show permission dialog or no

android/src/main/java/com/baseflow/permissionhandler/PermissionUtils.java
```diff
@@ -0,0 +1,31 @@
...
+public class PermissionUtils {
+
+  @RequiresApi(api = Build.VERSION_CODES.M)
+   static boolean neverAskAgainSelected(final Activity activity, final String permission) {
+    final boolean prevShouldShowStatus = getRationaleDisplayStatus(activity, permission);
+    final boolean currShouldShowStatus = ActivityCompat.shouldShowRequestPermissionRationale(activity, permission);
+    return prevShouldShowStatus != currShouldShowStatus;
+  }
+
+   static void setShouldShowStatus(final Context context, final String permission) {
```
mvanbeusekom: This method name suggests we are storing the outcome of the `shouldShowRequestPermissionRationale` which is not the case since you are simply storing the fact that the permissions have been requested. I suggest you rename this method to something like (note that this line also starts with one extra space that shouldn't be there):
```suggestion
  static void setRequestedPermission(final Context context, final String permission) {
```

android/src/main/java/com/baseflow/permissionhandler/PermissionUtils.java
```diff
@@ -0,0 +1,31 @@
...
+    SharedPreferences genPrefs = context.getSharedPreferences("GENERIC_PREFERENCES", Context.MODE_PRIVATE);
+    SharedPreferences.Editor editor = genPrefs.edit();
+    editor.putBoolean(permission, true);
+    editor.apply();
+  }
+
+  private static boolean getRationaleDisplayStatus(final Context context, final String permission) {
```
mvanbeusekom: This method name suggest that you are fetching a boolean indicating the current rationale display status, while in fact you are checking if the consuming App has requested the permission before. I suggest renaming the method to something like:
```suggestion
  private static boolean getRequestedPermissionBefore(final Context context, final String permission) {
```

android/src/main/java/com/baseflow/permissionhandler/PermissionUtils.java
```diff
@@ -0,0 +1,31 @@
...
+import android.os.Build;
+
+import androidx.annotation.RequiresApi;
+import androidx.core.app.ActivityCompat;
+
+public class PermissionUtils {
+
+  @RequiresApi(api = Build.VERSION_CODES.M)
+   static boolean neverAskAgainSelected(final Activity activity, final String permission) {
+    final boolean prevShouldShowStatus = getRationaleDisplayStatus(activity, permission);
```
mvanbeusekom: Variable name suggests it holds a result of a previous call to the `shouldShowRequestPermissionRationale`  while in fact it holds information if the App has requested the particular permission before. I suggest you rename the variable to something like:
```suggestion
    final boolean hasRequestedPermissionBefore = getRequestedPermissionBefore(activity, permission);
```

android/src/main/java/com/baseflow/permissionhandler/PermissionUtils.java
```diff
@@ -0,0 +1,31 @@
...
+import android.content.SharedPreferences;
+import android.os.Build;
+
+import androidx.annotation.RequiresApi;
+import androidx.core.app.ActivityCompat;
+
+public class PermissionUtils {
+
+  @RequiresApi(api = Build.VERSION_CODES.M)
+   static boolean neverAskAgainSelected(final Activity activity, final String permission) {
```
mvanbeusekom: The line seems to start with one space to much (it doesn't align correctly with the rest of the method block):

```suggestion
  static boolean neverAskAgainSelected(final Activity activity, final String permission) {
```

android/src/main/java/com/baseflow/permissionhandler/PermissionHandlerPlugin.java
```diff
@@ -204,11 +206,11 @@ private static int parseManifestName(String permission) {
...
-  private final Registrar mRegistrar;
-  private Result mResult;
-  private ArrayList<String> mRequestedPermissions;
-  @SuppressLint("UseSparseArrays")
-  private Map<Integer, Integer> mRequestResults = new HashMap<>();
+    private final Registrar mRegistrar;
+    private Result mResult;
+    private ArrayList<String> mRequestedPermissions;
+    @SuppressLint("UseSparseArrays")
+    private Map<Integer, Integer> mRequestResults = new HashMap<>();
```
mvanbeusekom: Remove added space at the beginning of the lines to fix alignment issue

mvanbeusekom:
#128

mvanbeusekom:
Hi @ManoyloK, just a quick update. Version 4.2.0 just went live containing your changes. Thanks again for your contribution.

PR #204, opened 2020-02-06
Request Location-Permission without active GPS

nibo:
I have made a few changes so that I can query the permission even without the gps active.

Perhaps this is a solution that interests others


mvanbeusekom:
I think this is a good idea, also other users have been requesting similar feature. This Android implementation should work, however we should also update this for the iOS side. 

After that we can also remove the `PermissionStatus.disabled` enum value.

PR #218, opened 2020-02-28
Feature/remove permissionstatus disabled

mvanbeusekom:

Feature

Unable to request location permissions when location services are disabled.

Allow users to request permissions while the location services are disabled.

Yes. The `PermissionStatus.disabled` has been removed (is of no use). Also renamed the `access_media_location` (to `accessMediaLocation`) and `activity_recognition` (to `activityRecognition`) permission groups to comply to the Effictive Dart naming conventions.

Run the example app with location services disabled.

- [x] All projects build

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,13 +1,8 @@
 # Flutter Permission handler Plugin
 
-[![pub package](https://img.shields.io/pub/v/permission_handler.svg)](https://pub.dartlang.org/packages/permission_handler)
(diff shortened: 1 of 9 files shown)


sososdk:
Hi @mvanbeusekom ,  I think android needs `disabled` to determine whether the device has turned off location services, to notify the user to turn on location services.

mvanbeusekom:
Hi @sososdk, the `permission_handler` plug-in has a separate method which can be used to determine if location services are enabled or not which can be used like this:

```
ServiceStatus serviceStatus = await PermissionHandler().checkServiceStatus(PermissionGroup.location);
```

We have only removed the `PermissionStatus.disabled` status which was part of the `requestPermissions` flow and actually blocking users from requesting permissions when location services was turned off.

sososdk:
Great! Thank you for your reply.

PR #289, opened 2020-05-13
iOS NotificationPermissionStrategy: Fixes wrong completionHandler call

phranck:
Bug fix

In case of iOS 10+ the call to grant permission was made outside of the `requestAuthorizationWithOptions(..)` block and returned immediately always with `PermissionStatusGranted`, which leads to a totally useless permission request.

This fix moves the call to the correct place - inside of `requestAuthorizationWithOptions(..)`'s block.

No.

- [x] All projects build

### DIFF
--- a/permission_handler/ios/Classes/strategies/NotificationPermissionStrategy.m
+++ b/permission_handler/ios/Classes/strategies/NotificationPermissionStrategy.m
@@ -37,7 +37,11 @@
           completionHandler(PermissionStatusDenied);
           return;
         }
+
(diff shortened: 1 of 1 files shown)


phranck:
> Great catch, thank you very much for the fix. Will release asap.

Wow, that was fast! 😍

mvanbeusekom:
Hehehe I happend to be behind the computer browsing issues when your PR came in ;). 

We just released version 5.0.0+hotfix.6 which has (only) your fix. Thanks again, much appreciated.

Issue #233, opened 2020-03-12
Unable to detect current Activity or App Context when executing from a Android service


ened:
## 🔙 Regression

The following error message is seen in the logs:
```
Unable to detect current Activity or App Context.
```

This seems to have been introduced by the new "neverAgain" checks which rely on a activity object to determine the permission status. This is causing problems for us because permission handler is running in a isolate that is started by an Android Service. Alas, no Activity object.

### Old (and correct) behavior

```
_permissionHandler.checkServiceStatus(p)
```
returns the correct value.

### Current behavior

```
_permissionHandler.checkServiceStatus(p)
```
now returns incorrect values.

### Configuration

**Version:** 4.4.0

**Platform:** 
- [ ] :iphone: iOS
- [x] :robot: Android

For the moment, we will revert back to permission handler 3.3.0.
In the meantime, I will start preparation on a PR to fix this issue, as it may have started around the time of the the V2 plugin migration.

ened:
In general the methods in `MethodCallHandlerImpl` should be split based on which methods need an activity and which ones does not. 

There is a problem with the neverAskAgain stuff, as it requires a activity to be figured out and it's currently part of the merged status for a permission. It may be enough to push that task to the developer and inform them that the results will be different based on the context they're running the Dart App in.

mvanbeusekom:
I am looking into the issue know, wouldn't it be as simple as moving to the `ApplicationContext`? At the moment we have easy access to both the `Activity` (if available) and the `ApplicationContext` (which is always available). 

So only when we need to show the dialog (e.g. when actually requesting permission) we would use the `Activity`. For all other actions we would use the `ApplicationContext`.

mvanbeusekom:
I also see the issue related to the `neverAskAgain` which indeed requires an Activity. I am thinking if we can work around this issue because.

ened:
The neverAskAgain check in https://github.com/Baseflow/flutter-permission-handler/blob/develop/permission_handler/android/src/main/java/com/baseflow/permissionhandler/PermissionUtils.java#L289 is the major pain point. Probably good idea to open a second bug report with Google. ;-)
Question is if that method could return a default value if `context` (is not a activity)...

mvanbeusekom:
I could make an extra check, if activity is `null` fallback to the ApplicationContext and ignore the `neverAskAgain` feature. This means however that you will only be informed about permissions being granted or denied when the app is running as an isolate. Would this be a workable solution?

[...]

PR #330, opened 2020-06-15
Relands bug fix for permission check in background

ened:

Bugfix

Can not check permission in BG service.

See https://github.com/Baseflow/flutter-permission-handler/pull/252 for a similar fix.

- [x] All projects build


ened:
This project could do with a extensive e2e test suite.

mvanbeusekom:
HI @ened, thanks again for supplying this bug fix. If I am not mistaken the real solution here is adding the `@Nullable` attribute to the `activity` parameter of the `determinePermissionStatus` method. I am wondering why did it go wrong in the first place, how come it was removed? As far as I can see it has never been there and on issue #233 you have confirmed the code was working correctly. 

I am trying to think how I can test changes better to prevent these kind of mistakes in the future. As you mentioned the project indeed could use a good e2e test suite and I have been playing around with it but unfortunately without a lot of success (I keep on running into error when running the `flutter drive` command). One thing that would be very helpful is we could somehow add some background functionality to the example app so we can at least run that and see if it fails. Would this be something you think is an option and maybe even something you are willing to help out with?

Besides this, the code in the PR looks good and I will merge and release it later this week.

ened:
The `@Nullable` annotation can be omitted, it's just there fore clarity.

The support library method to determine the "neverAskAgain" status internally checks if activity is null and if so just returns false, which means in the permission_handler code, the next else branch is called up.

Main change is the removal of the `activity` requirement to check for the status and this was solved in the same way in #233. I believe that issue was handling V4 of the permission_handler plugin.

Rgd. the E2E tests, something can be done, given the recent improvements in the `e2e` plugin. It is now possible run service/activity tests with permissions pre-granted, which is nice. The only issue is how to get the information on which native tests is being run at the moment up to the Dart e2e code. I'll try to update the simple e2e tests for this plugin in a separate PR.

mvanbeusekom:
I see now where you removed the `if(activity == null)` block, this indeed came back when we refactored the plugin into 5.0. The changes in issue #233 happened at the same time and I guess this wasn't merged in correctly when sifting through the merge conflicts.

mvanbeusekom:
And of course looking forward to your PR regarding the e2e tests. Even a simple setup might be helpful to get me started and to expand upon.

PR #389, opened 2020-09-14
Add draw overlay permission

niteshgrg:
Adds support for new SYSTEM_ALERT_WINDOW permission.

Currently we cannot ask for this permission.

Now developers can request for draw over apps permission and redirect to desired activity to give that.

No

- [x] All projects build


permission_handler/android/src/main/java/com/baseflow/permissionhandler/PermissionManager.java
```diff
@@ -120,6 +120,14 @@ void requestPermissions(
                 intent.setAction(Settings.ACTION_REQUEST_IGNORE_BATTERY_OPTIMIZATIONS);
                 intent.setData(Uri.parse("package:" + packageName));
                 activity.startActivityForResult(intent, PermissionConstants.PERMISSION_CODE_IGNORE_BATTERY_OPTIMIZATIONS);
+            } else if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.M && permission == PermissionConstants.PERMISSION_GROUP_SYSTEM_ALERT_WINDOW) {
+                activityRegistry.addListener(new ActivityResultListener(successCallback, activity));
```
mvanbeusekom, line 124: I think we should handle this differently. This can lead to memory leaks or unwanted behavior since we register a listener but never remove it. 

Current solution is to keep track of a boolean to indicate if the callback code has already ran for that particular instance of the result listener, but in the meantime the listener is still around. So if you request the permission multiple times during the live time of the App their will be multiple listeners in memory and all will be fired each time an activity result comes in.

It would be better to register a listener once in advance and not each time the permission is requested..

permission_handler/android/src/main/java/com/baseflow/permissionhandler/PermissionManager.java
```diff
@@ -271,21 +292,38 @@ private int checkNotificationPermissionStatus(Context context) {
...
+                        : PermissionConstants.PERMISSION_STATUS_DENIED;
+
+                HashMap<Integer, Integer> results = new HashMap<>();
+                results.put(PermissionConstants.PERMISSION_GROUP_IGNORE_BATTERY_OPTIMIZATIONS, status);
+                callback.onSuccess(results);
+                return true;
+            } else if (requestCode == PermissionConstants.PERMISSION_GROUP_SYSTEM_ALERT_WINDOW && Build.VERSION.SDK_INT >= Build.VERSION_CODES.M) {
+                if (alreadySystemAlertWindowCallbackCalled)
+                    return false;
+                alreadySystemAlertWindowCallbackCalled = true;
```
mvanbeusekom, line 317: As mentioned before, we should improve the registration of the activity listener so we don't need this construct and the `onActivityResult` method is only called once.

mvanbeusekom (changes requested):
I am having problems with the registration of the activity listener each time we request the `IGNORE_BATTERY_OPTIMIZATIONS` or `MANAGE_OVERLAY_PERMISSION` permissions. This will lead to memory leaks.

permission_handler/android/src/main/java/com/baseflow/permissionhandler/PermissionManager.java
niteshgrg: @mvanbeusekom One way i can think of is to increase boolean scope so when we register we set it to true. And we don't set the listener till its false. We set it back to false in result one. As if a user has come back and we trigger the intent again we would want to have listener attached to it.

mvanbeusekom: @niteshgrg, I recently faced a similar problem when implementing the geolocator plugin. Here I registered the listener on the plugin level (when the activity gets attached) and remove the listener when the activity is detached. Then I simply invoke a method on the class responsible of handling the activity result. 

This makes sure that listeners are only registered once and make communication pretty straight forward. Maybe you could do something similar here. If you want to have a look checkout the [GeolocatorPlugin](https://github.com/Baseflow/flutter-geolocator/blob/develop/geolocator/android/src/main/java/com/baseflow/geolocator/GeolocatorPlugin.java) class.

niteshgrg: Hey i actually haven't done native development before maybe thats why now able to solve the problem. I tried to go via the example you gave. In order to get ActivityListener in main class i was doing something like this but doesn't seem to work like this. Any appreciate any suggestions or advice -

```
public PermissionHandlerPlugin() {
        this.permissionManager = PermissionManager.ActivityResultListener();
    }
```

niteshgrg:
@mvanbeusekom made few changes to make it a global listener. Let me know If things look fine now. Thanks

jja08111:
Any update?

LukeDaniel16:
Any update?

mvanbeusekom:
Implemented in PR #529

PR #451, opened 2021-02-12
Nullsafety

danielroek:
Null safety

Not null safe

Null safe

No

- [x] All projects build

### DIFF
--- a/permission_handler/example/pubspec.yaml
+++ b/permission_handler/example/pubspec.yaml
@@ -2,7 +2,7 @@ name: permission_handler_example
 description: Demonstrates how to use the permission_handler plugin.
 
 environment:
-  sdk: ">=2.1.0 <3.0.0"
+  sdk: ">=2.12.0-0 <3.0.0"
 
 dependencies:
   flutter:
@@ -15,7 +15,7 @@ dev_dependencies:
   permission_handler:
     path: ../
 
-  url_launcher: ^5.4.11
+  url_launcher: ^6.0.0-nullsafety.6
 
 flutter:
   uses-material-design: true
--- a/permission_handler/example/pubspec.yaml
+++ b/permission_handler/example/pubspec.yaml
@@ -15,7 +15,7 @@ dev_dependencies:
   permission_handler:
     path: ../
 
-  url_launcher: ^6.0.0-nullsafety.6
+  url_launcher: ^6.0.0
 
 flutter:
   uses-material-design: true
--- a/permission_handler/example/pubspec.yaml
+++ b/permission_handler/example/pubspec.yaml
@@ -2,7 +2,7 @@ name: permission_handler_example
 description: Demonstrates how to use the permission_handler plugin.
 
 environment:
-  sdk: ">=2.12.0-0 <3.0.0"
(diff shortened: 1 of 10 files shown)


permission_handler/example/pubspec.yaml
```diff
@@ -2,7 +2,7 @@ name: permission_handler_example
 description: Demonstrates how to use the permission_handler plugin.
 
 environment:
-  sdk: ">=2.1.0 <3.0.0"
+  sdk: ">=2.12.0-0 <3.0.0"
```
mvanbeusekom, line 5: Update this to the NNBD stable version:

```suggestion
  sdk: ">=2.12.0-259.9.beta <3.0.0"
```

permission_handler/example/pubspec.yaml
```diff
@@ -15,7 +15,7 @@ dev_dependencies:
   permission_handler:
     path: ../
 
-  url_launcher: ^5.4.11
+  url_launcher: ^6.0.0-nullsafety.6
```
mvanbeusekom, line 18: Can be updated to the stable version:
```suggestion
  url_launcher: ^6.0.0
```

permission_handler/pubspec.yaml
```diff
@@ -1,6 +1,6 @@
 name: permission_handler
 description: Permission plugin for Flutter. This plugin provides a cross-platform (iOS, Android) API to request and check permissions.
-version: 5.1.0+2
+version: 6.0.0-nullsafety
```
mvanbeusekom, line 3: We can update directly to stable:
```suggestion
version: 6.0.0
```

permission_handler/pubspec.yaml
```diff
@@ -15,13 +15,13 @@ flutter:
...
 
 dev_dependencies:
-  effective_dart: ^1.2.1
-  plugin_platform_interface: ^1.0.1
+  effective_dart: ^1.3.0
+  plugin_platform_interface: ^1.0.3
 
 environment:
-  sdk: ">=2.7.0 <3.0.0"
+  sdk: ">=2.12.0-0 <3.0.0"
```
mvanbeusekom, line 26: Update this to the NNBD stable version:

```suggestion
  sdk: ">=2.12.0-259.9.beta <3.0.0"
```

permission_handler/lib/permission_handler.dart
```diff
@@ -46,7 +46,8 @@ extension PermissionActions on Permission {
   ///
   /// Returns the new [PermissionStatus].
   Future<PermissionStatus> request() async {
-    return (await [this].request())[this];
+    final permissionStatus = (await [this].request())[this];
+    return permissionStatus!;
```
mvanbeusekom, line 50: Is the bang (!) operator really necessary here? I would expect that the result from the `(await [this].request())[this]` call should never be null (I don't see any nullable object "?" notations).

If the `permissionStatus` variable is allowed to be null we might need to investigate and solve that instead of simply adding a bang operator (as this might still cause null exceptions on runtime since we cannot be sure it's always not null).

mvanbeusekom (changes requested):

mvanbeusekom (approved):
LGTM, but before we publish this we should also include your other PR #453

PR #453, opened 2021-02-19
Remove shared pref

danielroek:
Removes the use of SharedPreferences in the Android implementation. 

Shows undetermined on unknown permission status
Saves 'asked' in SharedPrefs

Shows denied instead of undetermined
Does not save 'asked' in SharedPrevs. 

Yes, undetermined is removed. 

Please test permanently denying permissions for correct behaviour 

[Geolocator's related issue](https://github.com/Baseflow/flutter-geolocator/wiki/Breaking-changes-in-7.0.0#android-permission-update
)

- [X] All projects build
- [X] Follows style guide lines ([code style guide](https://github.com/Baseflow/flutter-permission-handler/blob/develop/CONTRIBUTING.md))
- [X] Rebased onto current develop


permission_handler/android/src/main/java/com/baseflow/permissionhandler/PermissionUtils.java
```diff
@@ -279,20 +274,7 @@ static boolean isNeverAskAgainSelected(final Activity activity, final String nam
...
-    SharedPreferences.Editor editor = genPrefs.edit();
-    editor.putBoolean(permission, true);
-    editor.apply();
-  }
-
-  static boolean getRequestedPermissionBefore(final Context context, final String permission) {
-    SharedPreferences genPrefs = context.getSharedPreferences("GENERIC_PREFERENCES", Context.MODE_PRIVATE);
-    return genPrefs.getBoolean(permission, false);
+    return !shouldShowRequestPermissionRationale;
   }
```
mvanbeusekom, line 279: Lets merge this method together with the `isNeverAskAgainSelected` method and clean up code a little bit:

```java
    @RequiresApi(api = Build.VERSION_CODES.M)
    static boolean isNeverAskAgainSelected(final Activity activity, final String name) {
        if (activity == null) {
            return false;
        }
        
        final boolean shouldShowRequestPermissionRationale = ActivityCompat.shouldShowRequestPermissionRationale(activity, permission);
        return !shouldShowRequestPermissionRationale; 
    }
```

mvanbeusekom (changes requested):
I think this looks good in general just some small remarks that cleans up the code.

mvanbeusekom:
@danielroek this looks good. Since we are introducing breaking changes I think we should also remove the `notDetermined` permission status since that also results in a lot of problems (especially on the Android side). 

This means we need to also update the iOS side and map the `notDetermined` status as `denied`. Could you also include this in this PR? And would be great if we can document it as wel in the README.md.

permission_handler/ios/Classes/PermissionManager.m
```diff
@@ -75,8 +75,7 @@ + (void)openAppSettings:(FlutterResult)result {
         BOOL success = [[UIApplication sharedApplication] openURL:[NSURL URLWithString:UIApplicationOpenSettingsURLString]];
         result([[NSNumber alloc] initWithBool:success]);
     } else {
-        result(@false);
-    }
+        result(@false);    }
```
danielroek, line 78:
```suggestion
        result(@false);    
    }
```

mvanbeusekom:
Hi @danielroek, 

I receive the following error when on Android I permanently deny permissions:

```
RangeError (RangeError (index): Invalid value: Not in inclusive range 0..4: 5)
```

Looks like the Android code returns index number `5` while the `PermissionStatus` enum on the Dart side only contains indexes in the range `0..4`.

To reproduce this take the following steps:
- Run the example App;
- Request a permission;
- Select "Deny" from the permission dialog;
- Request the same position again;
- Check the "Don't ask again" checkbox in the permission dialog;
- Again select the "Deny" option;

mvanbeusekom (changes requested):
Hi @danielroek,

I receive the following error when on Android I permanently deny permissions:

```
RangeError (RangeError (index): Invalid value: Not in inclusive range 0..4: 5)
```

Looks like the Android code returns index number 5 while the PermissionStatus enum on the Dart side only contains indexes in the range 0..4.

To reproduce this take the following steps:

- Run the example App;
- Request a permission;
- Select "Deny" from the permission dialog;
- Request the same position again;
- Check the "Don't ask again" checkbox in the permission dialog;
- Again select the "Deny" option;

danielroek:
Updated the implementation. I made an error with the enum implementation that has been fixed now.

mvanbeusekom (approved):
Had to fix some permissions on iOS but looks good now.

PR #465, opened 2021-03-04
fix Error: The getter 'isUndetermined' isn't defined for the class 'Future<PermissionStatus>'

mo-ah-dawood:


mvanbeusekom:
Duplicate of #453

PR #469, opened 2021-03-05
Removed isUndetermined from PermissionCheckShortcuts

marcossevilla:

Bug fix, the `undetermined` permission status was used for `PermissionCheckShortcuts`'s `isUndetermined` extension method. Also, minor change in example for `PermissionStatus`, now it uses the `late` keyword for initialization, part of the null-safety migration.

There's an error in the `isUndetermined` method for the `PermissionCheckShortcuts` and the `PermissionStatus` on example is throwing an error because it initializes the `PermissionStatus` as `undetermined`, which is not defined.

`PermissionStatus` doesn't have any undetermined value, therefore you will have an error if you use that value in your apps.

Not a big one, you can say a deprecated property was removed.

Have Flutter 2 and Dart 2.12 installed.

- [ ✅] All projects build
- [✅ ] Follows style guide lines ([code style guide](https://github.com/Baseflow/flutter-permission-handler/blob/develop/CONTRIBUTING.md))
- [ ✅] Relevant documentation was updated
- [✅ ] Rebased onto current develop

### DIFF
--- a/permission_handler/example/lib/plugin_example/permission_widget.dart
+++ b/permission_handler/example/lib/plugin_example/permission_widget.dart
@@ -17,7 +17,7 @@ class _PermissionState extends State<PermissionWidget> {
   _PermissionState(this._permission);
 
   final Permission _permission;
-  PermissionStatus _permissionStatus = PermissionStatus.undetermined;
+  late PermissionStatus _permissionStatus;
 
   @override
   void initState() {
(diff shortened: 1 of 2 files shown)


mvanbeusekom:
Thank you for submitting this PR, but it is a duplicate of #453 which we are on the verge of merging.

PR #485, opened 2021-03-15
Added tests

JDDV:
Added tests for platform-handler

No

- [x] I updated pubspec.yaml with an appropriate new version according to the pub versioning philosophy.
- [x] I followed the style guide lines ([code style guide](https://github.com/Baseflow/flutter-permission-handler/blob/develop/CONTRIBUTING.md)).


permission_handler_platform_interface/lib/permission_handler_platform_interface.dart
```diff
@@ -1,6 +1,7 @@
 library permission_handler_platform_interface;
 
 import 'dart:async';
+import 'package:flutter/cupertino.dart';
```
mvanbeusekom, line 4: Do we need this import?

permission_handler_platform_interface/lib/src/permission_status.dart
```diff
@@ -3,10 +3,10 @@ part of permission_handler_platform_interface;
 /// Defines the state of a [Permission].
 enum PermissionStatus {
   /// The user granted access to the requested feature.
-  granted,
+  denied,
 
   /// The user denied access to the requested feature.
-  denied,
+  granted,
```
mvanbeusekom, line 9: The comments don't match the state anymore. These should be switched as well.

mvanbeusekom (changes requested):
Looks good but I have two small remarks.

permission_handler_platform_interface/lib/permission_handler_platform_interface.dart
JDDV: changed to meta/meta.dart import

mvanbeusekom (approved):
LGTM

Issue #491, opened 2021-03-18
MANAGE_EXTERNAL_STORAGE


aniketsongara:
Please add the support for Android 11 storage permission of 
`<uses-permission android:name="android.permission.MANAGE_EXTERNAL_STORAGE" />`

NebiSarigul:
++

JAICHANGPARK:
I Hope to add this permission

JAICHANGPARK:
I had some issues about permission manageExternalStorage
D/permissions_handler(  602): No permissions found in manifest for: 22

JDDV:
Hi @JAICHANGPARK, that's correct. In the platform_handler_platform_interface the permission is added, but not yet implemented in the permission_handler. I am currently working on it and it will be published probably this week. I'll give you an update as soon as it's published.

JAICHANGPARK:
@JDDV  Thank you for support ~!

JDDV:
@JAICHANGPARK, the manageExternalStorage is implemented now in the newest version of the plugin_handler (version 7.0.0).
I would like to point out that the documentation of using the manageExternalStorage permission says the following:

```
/// Android: Allows an application a broad access to external storage in
  /// scoped storage.
  /// iOS: Nothing
  ///
  /// You should request the Manage External Storage permission only when
  /// your app cannot effectively make use of the more privacy-friendly APIs.
  /// For more information: https://developer.android.com/training/data-storage/manage-all-files
  ///
  /// When the privacy-friendly APIs (i.e. [Storage Access Framework](https://developer.android.com/guide/topics/providers/document-provider)
  /// or the [MediaStore](https://developer.android.com/training/data-storage/shared/media) APIs) is all your app needs the
  /// [PermissionGroup.storage] are the only permissions you need to request.
  ///
  /// If the usage of the Manage External Storage permission is needed,
  /// you have to fill out the Permission Declaration Form upon submitting
  /// your app to the Google Play Store. More details can be found here:
  /// https://support.google.com/googleplay/android-developer/answer/9214102#zippy=
```

So take in good consideration if the permission really is needed. If it's not needed, as mentioned above, you can use the PermissionGroup.storage which is already implemented.

I hope this will help you out!

Ch4rl3B:
> 
> 
> I had some issues about permission manageExternalStorage
> D/permissions_handler( 602): No permissions found in manifest for: 22

I have this issue.. I'm using permission_handler: ^7.1.0
[...]

PR #526, opened 2021-04-21
Added MANAGE_EXTERNAL_STORAGE permission

JDDV:

1. Registered listeners on the plugin level to prevent memory leaks or unwanted behaviour
2. Added MANAGE_EXTERNAL_STORAGE permissions for Android 11 and higher

When requesting a permission multiple times during the live time of the App there will be multiple listeners in memory and all will be fired each time an activity result comes in

The listeners get registered on the plugin level so there will not be multiple listeners put in memory when requesting multiple permissions.

Also added the MANAGE_EXTERNAL_STORAGE permission for Android 11 and up.

Yes, build.gradle needed to be updated, compile SDK version is now set to 30 in the build.gradle for handling the MANAGE_EXTERNAL_STORAGE permission

Listeners bug mentioned in #389 will be solved
Added MANAGE_EXTERNAL_STORAGE as requested in #491  

### DIFF
--- a/permission_handler/android/src/main/java/com/baseflow/permissionhandler/PermissionManager.java
+++ b/permission_handler/android/src/main/java/com/baseflow/permissionhandler/PermissionManager.java
@@ -226,17 +226,13 @@ final class PermissionManager implements PluginRegistry.ActivityResultListener, 
             }
 
             if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.M && permission == PermissionConstants.PERMISSION_GROUP_IGNORE_BATTERY_OPTIMIZATIONS) {
-                String packageName = activity.getPackageName();
-                Intent intent = new Intent();
-                intent.setAction(Settings.ACTION_REQUEST_IGNORE_BATTERY_OPTIMIZATIONS);
-                intent.setData(Uri.parse("package:" + packageName));
-                activity.startActivityForResult(intent, PermissionConstants.PERMISSION_CODE_IGNORE_BATTERY_OPTIMIZATIONS);
+                executeIntent(
+                        Settings.ACTION_REQUEST_IGNORE_BATTERY_OPTIMIZATIONS,
+                        PermissionConstants.PERMISSION_CODE_IGNORE_BATTERY_OPTIMIZATIONS);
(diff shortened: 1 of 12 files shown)


permission_handler/android/src/main/java/com/baseflow/permissionhandler/PermissionManager.java
```diff
@@ -111,38 +213,37 @@ void requestPermissions(
...
                 Intent intent = new Intent();
                 intent.setAction(Settings.ACTION_REQUEST_IGNORE_BATTERY_OPTIMIZATIONS);
                 intent.setData(Uri.parse("package:" + packageName));
                 activity.startActivityForResult(intent, PermissionConstants.PERMISSION_CODE_IGNORE_BATTERY_OPTIMIZATIONS);
+            } else if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.R && permission == PermissionConstants.PERMISSION_GROUP_MANAGE_EXTERNAL_STORAGE) {
+                String packageName = activity.getPackageName();
+                Intent intent = new Intent();
+                intent.setAction(Settings.ACTION_MANAGE_APP_ALL_FILES_ACCESS_PERMISSION);
+                intent.setData(Uri.parse("package:" + packageName));
+                activity.startActivityForResult(intent, PermissionConstants.PERMISSION_CODE_MANAGE_EXTERNAL_STORAGE);
```
mvanbeusekom, line 239: It would be good to extract this code into a separate method to avoid repeating code. I am thinking something like:

```java
private void executeIntent(int action, int requestCode) {
    String packageName = activity.getPackageName();
    Intent intent = new Intent();
    intent.setAction(action);
    intent.setData(Uri.parse("package:" + packageName));
    activity.startActivityForResult(intent, requestCode);
}
```

And then we can call it something like this:
```suggestion
                executeIntent(
                    Settings.ACTION_MANAGE_APP_ALL_FILES_ACCESS_PERMISSION, 
                    PermissionConstants.PERMISSION_CODE_MANAGE_EXTERNAL_STORAGE);
```

mvanbeusekom (changes requested):
Look good, just one small remark to reduce code duplication

permission_handler/android/src/main/java/com/baseflow/permissionhandler/PermissionManager.java
JDDV: Good call, change has been made

mvanbeusekom (approved):

PR #662, opened 2021-09-03
Android 12 (API 31) deprecation error fix when building appbundle

florissmit1:
Bug fix

Currently when the `compileSdkVersion` is set to 31, the `flutter build appbundle` will not succeed due to deprecation warnings/errors. 

When using API level 15 or above, BluetoothManagert.getAdapter() is used, otherwise the (deprecated) method BluetoothAdapter.getDefaultAdapter() is used. The method will continue to work, but  developers are strongly encouraged to migrate to using BluetoothManager#getAdapter(), since that approach enables support for Context#createAttributionContext.

No

Does not apply

https://developer.android.com/reference/android/bluetooth/BluetoothAdapter.html#getDefaultAdapter()

### DIFF
--- a/permission_handler/android/src/main/java/com/baseflow/permissionhandler/ServiceManager.java
+++ b/permission_handler/android/src/main/java/com/baseflow/permissionhandler/ServiceManager.java
@@ -1,6 +1,9 @@
 package com.baseflow.permissionhandler;
 
+import static android.content.Context.BLUETOOTH_SERVICE;
+
 import android.bluetooth.BluetoothAdapter;
+import android.bluetooth.BluetoothManager;
 import android.content.Context;
 import android.content.Intent;
 import android.content.pm.PackageManager;
@@ -45,7 +48,7 @@ final class ServiceManager {
             return;
         }
         if(permission == PermissionConstants.PERMISSION_GROUP_BLUETOOTH){
-            final int serviceStatus = isBluetoothServiceEnabled()
+            final int serviceStatus = isBluetoothServiceEnabled(context)
                     ? PermissionConstants.SERVICE_STATUS_ENABLED
                     : PermissionConstants.SERVICE_STATUS_DISABLED;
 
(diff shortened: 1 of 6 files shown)


chiragabhani (approved):

radhen-thirstydevs (approved):

DarkPhantom146 (approved):

permission_handler/android/src/main/java/com/baseflow/permissionhandler/ServiceManager.java
```diff
@@ -147,9 +150,16 @@ private static boolean isLocationServiceEnablePreKitKat(Context context)
...
+    // pre S versions of Android
+    @SuppressWarnings("deprecation")
+    private boolean isBluetoothServiceEnabled(Context context) {
+        if (android.os.Build.VERSION.SDK_INT >= android.os.Build.VERSION_CODES.JELLY_BEAN_MR2) {
+            BluetoothManager manager = (BluetoothManager) context.getSystemService(BLUETOOTH_SERVICE);
+            final BluetoothAdapter adapter = manager.getAdapter();
+            return adapter.isEnabled();
+        } else {
+            return BluetoothAdapter.getDefaultAdapter().isEnabled();
+        }
```
mvanbeusekom, line 163: I feel we should reverse this `if` statement. So that the default behaviour would be to use the non-deprecated API and only use the deprecated API when the SDK version is below `JELLY_BEAN_MR2`:

```suggestion
        if (android.os.Build.VERSION.SDK_INT < android.os.Build.VERSION_CODES.JELLY_BEAN_MR2) {
            return BluetoothAdapter.getDefaultAdapter().isEnabled();
        } 
        
        BluetoothManager manager = (BluetoothManager) context.getSystemService(BLUETOOTH_SERVICE);
        final BluetoothAdapter adapter = manager.getAdapter();
        return adapter.isEnabled();
```

mvanbeusekom (changes requested):
Default behaviour should be the non-deprecated code path.

mvanbeusekom (approved):

PR #822, opened 2022-03-28
Add Persistent flag for permanentlyDenied LocationAlways permission

lukaskurz:

Its a bug fix for requesting locationAlways when it is permanently denied.

When the locationAlways permission gets denied once, it is permanentlyDenied by the system (iOS), meaning future requests for the permission are ignored. This leads to a dangling callback that is never called if the permission was previously permanentlyDenied. This in turn causes the async function in dart to get stuck and never fulfil, as well as future permission calls will throw the `ERROR_ALREADY_REQUESTING_PERMISSIONS` error.

Once the locationAlways permission gets requested, a flag is set in the persistent storage in `UserDefaults`. This way the library knows that is has already used its 1 request that it is allowed. The next time the permission is requested, the flag is checked, and if it is set and true, then we can use the status to determine whether we have permission or were permanentlyDenied. Either way, we don't try to request, once we know, we can no longer send permission requests for locationAlways, preventing the dangling callback.

Not really.

@mvanbeusekom mentioned this change would be welcome

https://github.com/Baseflow/flutter-permission-handler/pull/800#issuecomment-1068835410

- [X] I made sure all projects build.
- [X] I updated pubspec.yaml with an appropriate new version according to the [pub versioning philosophy](https://dart.dev/tools/pub/versioning).
- [X] I updated CHANGELOG.md to add a description of the change.
- [X] I followed the style guide lines ([code style guide](https://github.com/Baseflow/flutter-permission-handler/blob/master/CONTRIBUTING.md)).
- [X] I updated the relevant documentation.
- [X] I rebased onto current `master`.


mvanbeusekom (approved):
LGTM, thank you for this contribution

mvanbeusekom:
Hi @lukaskurz,

After some investigation it turns out the GitHub no longer allows accessing a repo through the git protocol (git://) in an unauthenticated way and instead suggests to use the HTTP protocol. Turns out all our example apps have a dependency on a Baseflow template which we access via a git reference (using the git protocol) and this was blocking / failing the build process.

I have just merged an update, could you please rebase your branch on the latest master and push the update? After that the build should run fine again. If everything turns green (which I have no doubt it will) I will merge the PR and publish a new version.

Thanks and apologies for the inconvenience.

lukaskurz:
@mvanbeusekom rebase should be done 👍

Issue #917, opened 2022-09-20
locationAlways does not work on Android 10


hirameee55:
## 🔙 Regression
When requesting locationAlways on Android 10, the Runtime Permission dialog does not appear and the API is returned immediately.

According to the Android specification, the "Allow all the time" option is included even when requesting background location on Android 10.
https://developer.android.com/training/location/permissions#request-background-location

### Old (and correct) behavior
Version:8.3.0
When I request locationAlways on Android 10, the Runtime Permission dialog appears and I can select "Allow all the time".

### Current behavior
Version: 10.0.0
Runtime Permission does not appear on Android 10.

### Reproduction steps
Do the following on Android 10 device.

`await Permission.locationAlways.request();`

### Configuration

**Version:** 10.0.0

**Platform:** 
- [ ] :iphone: iOS
- [x] :robot: Android

**flutter doctor:** 
[✓] Flutter (Channel stable, 3.0.5, on macOS 12.5.1 21G83 darwin-x64, locale en-JP)
[✓] Android toolchain - develop for Android devices (Android SDK version 33.0.0)
[✓] Xcode - develop for iOS and macOS (Xcode 14.0)
[✓] Chrome - develop for the web
[✓] Android Studio (version 2020.3)
[✓] VS Code (version 1.71.2)
[✓] Connected device (3 available)
[✓] HTTP Host Availability

Erhannis:
MIGHT be relevant: on Android 12, at least, I was asking for multiple permissions and it was failing without explanation, but after half an hour of poking around, I looked at logcat and it said "GrantPermissionsViewModel: For R+ apps, background permissions must be requested after foreground permissions are already granted".  Separating out the "locationAlways" into its own subsequent request fixed it.  (IMO, the library should do that automatically, please.)

hirameee55:
This is true for Android R (11) and above, but Android Q (10) can request background and foreground permissions at the same time. The following modification makes this impossible and requires the addition of "Android R or higher" to the break requirement.
https://github.com/Baseflow/flutter-permission-handler/pull/792/files
https://github.com/Baseflow/flutter-permission-handler/pull/794/files

okbye123:
This might help https://www.youtube.com/watch?v=uMvGpBOT0ZY

JeroenWeener:
Related to #781. This issue requires logic specific for Android 10.

PR #794, opened 2022-02-16
Issue 793

emarwil-cr:
### :sparkles: What kind of change does this PR introduce? (Bug fix, feature, docs update...)
This change fixes the regression described in Issue #793

### :arrow_heading_down: What is the current behavior?
Current behavior prevents Android 9 (Pie) or earlier from having locationAlways permission granted due to a misplaced 'break' statement.

### :new: What is the new behavior (if this is a feature change)?
Behavior is restored to how it worked in v8.3.0 for Android 9 (Pie) and earlier.

### :boom: Does this PR introduce a breaking change?
No.  Compared location permission behavior on Android 12 for both 9.0.1 and the current PR.  Behavior is the same.

### :bug: Recommendations for testing
Follow steps described in Issue #793 to confirm the behavior is now the same as it was in version 8.3.0 for location permissions.

### :memo: Links to relevant issues/docs
Issue #793: https://github.com/Baseflow/flutter-permission-handler/issues/793

### :thinking: Checklist before submitting

- [x] I made sure all projects build.
- [x] I updated pubspec.yaml with an appropriate new version according to the [pub versioning philosophy](https://dart.dev/tools/pub/versioning).
- [x] I updated CHANGELOG.md to add a description of the change.
- [x] I followed the style guide lines ([code style guide](https://github.com/Baseflow/flutter-permission-handler/blob/master/CONTRIBUTING.md)).
- [x] I updated the relevant documentation.
- [x] I rebased onto current `master`.

PR #921, opened 2022-09-23
Fix locationAlways does not work on Android 10

hirameee55:
This change fixes the regression described in Issue #917.

The current behavior is that Runtime Permission is not shown when requesting locationAlways on Android 10 (Q).

In Android Q (10), it is possible to request background and foreground permissions at the same time. LocationAlways satisfies this.
Android 11 (R) or higher cannot request background and foreground permissions at the same time, so v10.0.1 behavior is not affected.
These are Android specifications.
https://developer.android.com/training/location/permissions#request-background-location

No.

When I request locationAlways on Android 10 (Q), the Runtime Permission appears and I can select "Allow all the time".

- [X] I made sure all projects build.
- [X] I updated pubspec.yaml with an appropriate new version according to the [pub versioning philosophy](https://dart.dev/tools/pub/versioning).
- [X] I updated CHANGELOG.md to add a description of the change.
- [X] I followed the style guide lines ([code style guide](https://github.com/Baseflow/flutter-permission-handler/blob/master/CONTRIBUTING.md)).
- [X] I updated the relevant documentation.
- [X] I rebased onto current `master`.

### DIFF
--- a/permission_handler_android/android/src/main/java/com/baseflow/permissionhandler/PermissionUtils.java
+++ b/permission_handler_android/android/src/main/java/com/baseflow/permissionhandler/PermissionUtils.java
@@ -117,7 +117,9 @@ public class PermissionUtils {
                 if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.Q) {
                     if (hasPermissionInManifest(context, permissionNames, Manifest.permission.ACCESS_BACKGROUND_LOCATION))
                         permissionNames.add(Manifest.permission.ACCESS_BACKGROUND_LOCATION);
-                    break;
+
+                    if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.R)
+                        break;
                 }
             case PermissionConstants.PERMISSION_GROUP_LOCATION_WHEN_IN_USE:
             case PermissionConstants.PERMISSION_GROUP_LOCATION:
--- a/permission_handler_android/CHANGELOG.md
+++ b/permission_handler_android/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 10.0.1
+
+* Fix locationAlways does not work on Android 10.
+
 ## 10.0.0
 
  * __BREAKING CHANGE__: Updated Android `compileSdkVersion` to `33` to handle the new `POST_NOTIFICATIONS` permission.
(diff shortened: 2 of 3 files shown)


hirameee55:
@mvanbeusekom 
I need the approval of the maintainer to have it reviewed.

permission_handler_android/android/src/main/java/com/baseflow/permissionhandler/PermissionUtils.java
```diff
@@ -117,7 +117,9 @@ static List<String> getManifestNames(Context context, @PermissionConstants.Permi
                 if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.Q) {
                     if (hasPermissionInManifest(context, permissionNames, Manifest.permission.ACCESS_BACKGROUND_LOCATION))
                         permissionNames.add(Manifest.permission.ACCESS_BACKGROUND_LOCATION);
-                    break;
+
+                    if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.R)
+                        break;
```
mvanbeusekom, line 122: Reading through the [documentation](https://developer.android.com/training/location/permissions), I think we should always fall through to the location case and add the `ACCESS_COARSE_LOCATION` and/ or `ACCESS_FINE_LOCATION` as well. Meaning we need to remove the `break;` statement al together. 

We should also improve the comment above to reflect this change.

mvanbeusekom (changes requested):

permission_handler_android/android/src/main/java/com/baseflow/permissionhandler/PermissionUtils.java
hirameee55: Hi @mvanbeusekom 
Thank you for your review.

If remove the `break;`, #794 doesn't work.
Android R (11) and above cannot request background and foreground permissions at the same time.
Therefore, for Android R (11) and above, it is ideal to break if `Manifest.permission.ACCESS_BACKGROUND_LOCATION` is included and acquire permissions only in the background.

hirameee55: Hi @mvanbeusekom 
Could you confirm my comments?
Please feel free to contact us if you have any questions.

Issue #988, opened 2023-02-08
Feature Request: iOS Provisional Notification Permissions


azack:
## 🚀 Feature Requests

* Add support for distinguishing the iOS `provisional` notification permission for checking current permission level & requesting full permission

### Contextualize the feature

In iOS 12, [UNAuthorizationStatusProvisional](https://developer.apple.com/documentation/usernotifications/unauthorizationstatus/unauthorizationstatusprovisional?language=objc) was added as a case to [UNAuthorizationStatus](https://developer.apple.com/documentation/usernotifications/unauthorizationstatus?language=objc).  This allows a developer to request notification permissions in a way that are initially approved without the user being prompted on a trial basis, but with the restriction that users are not actively interrupted when a notification is received.  It is up to the user to check Notification Center themselves and decide if they want to keep the notifications.  Much more background in Apple's article [Asking permission to use notifications](https://developer.apple.com/documentation/usernotifications/asking_permission_to_use_notifications).

Currently, if an application has already been granted provisional notification permissions, retrieving the permission via `permission_handler`'s [NotificationPermissionStrategy](https://github.com/Baseflow/flutter-permission-handler/blob/2e47f05ae9dabee2bc671166301aa948259b39f2/permission_handler_apple/ios/Classes/strategies/NotificationPermissionStrategy.m#L61-L81) `permissionStatus` method will return `PermissionStatusGranted`.   This does not reflect the actual permission, which would cause client code to think permission is fully granted.  If client code wanted to "upgrade" a user's permission to fully authorized by prompting them to enable this permission, we can't, because [requestPermission:](https://github.com/Baseflow/flutter-permission-handler/blob/2e47f05ae9dabee2bc671166301aa948259b39f2/permission_handler_apple/ios/Classes/strategies/NotificationPermissionStrategy.m#L22-L23) checks the permission status first and thinks it is already granted, showing no prompt.   By adding support for provisional access, client code could correctly trigger the OS permission dialog when explicit `authorized` access is desired.

If there is a downside to adding this, it's that changing the `permission_handler_platform_interface` may be unavoidable to support this additional permission type.  It is unclear to me if this could be done straightforwardly without a breaking change / a major version update; adding to the enum would mean that consumers of the api handling `PermissionStatus` in a `switch` statement could potentially require upgrading.   

[...]

PR #990, opened 2023-02-09
Issue #988: Provisional Notification Permissions Part 1/3

azack:

Implements the `platform_interface` part of #988 

* This package is not aware of the provisional status of notifications on iOS 12+.  

- [x] Adds the `provisional` permission status so that it can be used in a followup PR.  Split here so the package versioning works out.
- [x] Updates tests
- [x] Minor doc typo fix 

* It does introduce a new enum value which would be a breaking change for client code that switches on the enum and does not have a default case.  However, given that this is very small and because existing code is very likely running with a `pubspec.lock`, it should have limited impact.

* No direct testing impact in this changeset, though there will be in followup PR that adds detection for the new status.

#988 

### DIFF
--- a/permission_handler_platform_interface/CHANGELOG.md
+++ b/permission_handler_platform_interface/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 3.10.0
+
+* Added support for iOS 12+ Provisional Notification permissions.
+
 ## 3.9.0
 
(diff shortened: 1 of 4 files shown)


mvanbeusekom (approved):
LGTM, thank you for this contribution.

PR #991, opened 2023-02-09
Issue #988: Provisional Notification Permissions Part 2/3

azack:

Adds support to iOS 12+ Provisional notifications to `permission_handler_apple` as part of #988.

The code is not aware of the provisional status; existing apps with provisional status will return `granted` incorrectly, and requests for permission in a provisional state will be a noop instead of showing the user the permission dialog.

* This status is now handled

Not directly, but it depends on #990 which may be.

* Testing on all platforms should be unchanged, with the following exceptions:
*  apps with existing `provisional` notification status (e.g. notifications were requested with 
```swift
let options: UNAuthorizationOptions = [.alert, .sound, .badge, .provisional]
center.requestAuthorization(options: options)
```
For these apps, requesting permission via `permission_handler` should now result in the OS system notification permission dialog appearing when `final status = await perm.Permission.notification.request();` is made.
* apps with existing `provisional` notification permission status should return `provisional` from ` final status = await perm.Permission.notification.status;`

#988 

- [x] I made sure all projects build. **Note** requires the previous PR to be merged as the package is a dependency. Tested locally with `path: ` references.


mvanbeusekom (approved):
LGTM

Issue #884, opened 2022-08-01
[Android] Bluetooth permission required twice


ChristianNavolskyi:
## 🐛 Bug Report

I only want to test whether Bluetooth is turned on or off (with `flutter_blue_plus), therefore I would assume that only `<uses-permission android:name="android.permission.BLUETOOTH"/>` would be sufficient.

However, when I request Bluetooth permission I get `D/permissions_handler(27550): Bluetooth permission missing in manifest`.
The same is true when there is no `uses-permission` entry at all.

Setting the line twice like this however works:
```
<uses-permission android:name="android.permission.BLUETOOTH"/>
<uses-permission android:name="android.permission.BLUETOOTH"/>
```

This only happens in combination with `flutter_blue_plus`, but since the error message is reported by this plugin I wanted to start here.

I actually can request the bluetooth state even without the permission at all on my device (android 12) but I guess it's necessary on older devices.

### Expected behaviour

Requesting/getting the permission result without error messages.

### Reproduction steps

I created a small reproducing project [here](https://github.com/ChristianNavolskyi/permission_handler_bug).
In the `AndroidManifest.xml` the Bluetooth permission is only added once, which leads to the error above.

### Configuration

Phone: Pixel 5 with Android 12

**Version:** 10.0.0

**Platform:** 
- [ ] :iphone: iOS
- [x] :robot: Android

florissmit1:
Android 12's permission handling has changed, so the `<uses-permission android:name="android.permission.bluetooth>` must be changed, the permission is only used for devices which target API level 30 and lower. You can read more about it here; https://developer.android.com/guide/topics/connectivity/bluetooth/permissions.

I think this should be changed in the documentation of the permission handler, since the documentation is still targeting Android 11 and lower.

Erhannis:
I've encountered a similar problem, but with bluetooth AND location.  After duplicating the `BLUETOOTH` permission, I got an error "No permissions found in manifest for: []3", which number, looking at the code, maps to the location group.  I then duplicated my existing location permissions, and the app then successfully asked for permissions and ran correctly.  This is pretty clearly a bug in _something_; having the permissions present once should be sufficient, and having them present MORE than once should do nothing.

Permissions:
```
    <uses-permission android:name="android.permission.INTERNET"/>

    <uses-permission android:name="android.permission.BLUETOOTH" />
    <uses-permission android:name="android.permission.BLUETOOTH_ADMIN" />
[...]

PR #993, opened 2023-02-10
Fix android bluetooth permission request above api version 30

rufman:
The BLUETOOTH permission was removed in api version 30 in favor of BLUETOOTH_SCAN, BLUETOOTH_ADVERTISE and BLUETOOTH_CONNECT.

This means that manifests targetting api version above 30 will not have this permission set. For these versions the manifest should be checked for the new permissions.

This change ensures that the documented behavior of the bluetooth permission always returning `true` is true.

Fixes issues #884

Bug fix for Android api versions above 30 requesting Bluetooth permissions.

`Permission.bluetooth` return denied instead of the documented always allowed (https://pub.dev/documentation/permission_handler_platform_interface/latest/permission_handler_platform_interface/Permission/bluetooth-constant.html)

This commit brings the code in line with the documented behavior

No.

Test on api versions after 30 with a manifest that only includes one of the new permissions, and either no BLUETOOTH permission, or `<uses-permission android:name="android.permission.BLUETOOTH" android:maxSdkVersion="30" />

- https://pub.dev/documentation/permission_handler_platform_interface/latest/permission_handler_platform_interface/Permission/bluetooth-constant.html
- https://github.com/Baseflow/flutter-permission-handler/issues/884

### DIFF
--- a/permission_handler_android/android/src/main/java/com/baseflow/permissionhandler/PermissionManager.java
+++ b/permission_handler_android/android/src/main/java/com/baseflow/permissionhandler/PermissionManager.java
@@ -481,12 +481,36 @@ final class PermissionManager implements PluginRegistry.ActivityResultListener, 
     }
 
     private int checkBluetoothPermissionStatus(Context context) {
-        List<String> names = PermissionUtils.getManifestNames(context, PermissionConstants.PERMISSION_GROUP_BLUETOOTH);
-        boolean missingInManifest = names == null || names.isEmpty();
-        if (missingInManifest) {
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.S) {
+            // BLUETOOTH permission was removed in Android S in favor of BLUETOOTH_SCAN,
+            // BLUETOOTH_ADVERTISE and BLUETOOTH_CONNECT.
+            // This means above version 30 we need to check if any one of those permissions
+            // are present instead.
+            boolean scanPermission = checkPermissionStatus(context,
+                    PermissionConstants.PERMISSION_GROUP_BLUETOOTH_SCAN);
+            boolean advertisePermission = checkPermissionStatus(context,
+                    PermissionConstants.PERMISSION_GROUP_BLUETOOTH_ADVERTISE);
+            boolean connectPermission = checkPermissionStatus(context,
+                    PermissionConstants.PERMISSION_GROUP_BLUETOOTH_CONNECT);
+
+            if (!scanPermission && !advertisePermission && !connectPermission) {
+                Log.d(PermissionConstants.LOG_TAG, "Of the bluetooth permissions (BLUETOOTH_SCAN, BLUETOOTH_ADVERTISE or BLUETOOTH_CONNECT) missing in manifest");
+                return PermissionConstants.PERMISSION_STATUS_DENIED;
+            }
+            return PermissionConstants.PERMISSION_STATUS_GRANTED;
+        }
+        // legacy check for BLUETOOTH permission
+        boolean bluetoothPermission = checkPermissionStatus(context, PermissionConstants.PERMISSION_GROUP_BLUETOOTH);
+        if (!bluetoothPermission) {
             Log.d(PermissionConstants.LOG_TAG, "Bluetooth permission missing in manifest");
             return PermissionConstants.PERMISSION_STATUS_DENIED;
         }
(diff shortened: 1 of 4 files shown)


kismet- (approved):

permission_handler_android/android/src/main/java/com/baseflow/permissionhandler/PermissionManager.java
```diff
@@ -481,12 +481,36 @@ private int checkNotificationPermissionStatus(Context context) {
...
         }
         return PermissionConstants.PERMISSION_STATUS_GRANTED;
     }
+
+    private boolean checkPermissionStatus(Context context, @PermissionConstants.PermissionGroup int permission) {
+        List<String> names = PermissionUtils.getManifestNames(context, permission);
+        return names != null || !names.isEmpty();
+    }
+}
 }
```
KiliCode, line 516: Only one of these should remain. If both are kept this results in an error

C-8:
Is there an update on this? 
Right now this workaroung https://github.com/Baseflow/flutter-permission-handler/issues/884 is not even working anymore, since it throws error `Element uses-permission#android.permission.BLUETOOTH at AndroidManifest.xml:6:3-66 duplicated with element declared at AndroidManifest.xml:3:3-5:36`

permission_handler_android/android/src/main/java/com/baseflow/permissionhandler/PermissionManager.java
```diff
@@ -481,12 +481,36 @@ private int checkNotificationPermissionStatus(Context context) {
     }
 
     private int checkBluetoothPermissionStatus(Context context) {
-        List<String> names = PermissionUtils.getManifestNames(context, PermissionConstants.PERMISSION_GROUP_BLUETOOTH);
-        boolean missingInManifest = names == null || names.isEmpty();
-        if (missingInManifest) {
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.S) {
```
wujek-srujek, line 484: I have just spent some time with this library and getting to know its code and testing it on various Android devices, and here is the result of my research: this change (and PR) seems to be unnecessary, at least with version 10.4.2. 

As you can see here https://github.com/Baseflow/flutter-permission-handler/blob/permission_handler_v10.4.2/permission_handler_android/android/src/main/java/com/baseflow/permissionhandler/PermissionManager.java#L322-L324 the `checkBluetoothPermissionStatus` method edited here is only used if the new permissions are checked on a device with API level < S (31) - in this case the library will simply fall back to checking for the old `Permissions.BLUETOOTH`. If the current device is 31 or higher, the method will not be called and the permissions will be checked by the code starting here: https://github.com/Baseflow/flutter-permission-handler/blob/permission_handler_v10.4.2/permission_handler_android/android/src/main/java/com/baseflow/permissionhandler/PermissionManager.java#L327.

Similarly, `requestPermissions` (not touched in this PR) works correctly - on pre-31 devices it will first determine the status of the permissions (by using `checkBluetoothPermissionStatus` and the same fallback logic as described above) and because `Permissions.BLUETOOTH` is a 'normal' permission, as long as it is in the manifest it will be 'granted' (not having it in the manifest is an error), so requesting it is a no-op - it will not put the permission to `permissionsToRequest` list. On 31 and later devices, it will correctly request the new permissions.

Which means - as long as the Flutter part of you application uses new permissions, it will all work even on pre-31 devices as the library is smart enough to fallback to the old permission if new ones are used on new devices. Tested with both a 33 and 23 device.

As a side note, irrelevant to this PR, the same logic applies to the new `Permissions.POST_NOTIFICATIONS` introduced in API level 33 - if used on a pre-33 device, this library falls back to using the `NotificationManagerCompat` for the check, and 33 and later check the permission - see https://github.com/Baseflow/flutter-permission-handler/blob/permission_handler_v10.4.2/permission_handler_android/android/src/main/java/com/baseflow/permissionhandler/PermissionManager.java#L479-L492.

wujek-srujek:
> Is there an update on this? Right now this workaround #884 is not even working anymore, since it throws error `Element uses-permission#android.permission.BLUETOOTH at AndroidManifest.xml:6:3-66 duplicated with element declared at AndroidManifest.xml:3:3-5:36`

I have a test app running on devices with API level 23 and 33 and I don't need any workarounds. I always use the new permissions, it works even on old devices (see my other rather lengthy comment) and the lib implements fallbacks. I think this PR should be closed as it is unnecessary.

PR #1012, opened 2023-03-24
Added missing `return` statement on iOS.

OctavianLfrd:
Bug fix on iOS

The current behaviour is that if there is already an ongoing permission request, flutter `result` for the new request will be called twice, first will `FlutterError` and then with an actual request result. It is also possible that the previous result will never be called because the new result is by mistake stored in `_methodResult` property.

-

No

Try to perform multiple permission requests simultaneously.

-


mvanbeusekom:
Thank you for submitting this pull-request. We have just released version 9.0.8 which also contains this fix, therefore I am going to close this PR.

PR #1082, opened 2023-06-30
Fixes hang when requesting location always permission.

### LINKED ISSUE #802: How to handle situation if user chose "Allow Only Once" ?
Hi all, I would like to request permission for "location always" in my app. I first request permission for "location while in use", then if it's granted, I'll request "location always". 

However, in iOS 15, if the user chose "Allow Only Once" for "location while in use" permission, the permission status returns granted and the request for "location always" is stuck. I suspect the app could not request for "location always" when "only once" permission is granted. Now my app is stuck at the black loading screen. How could I detect if the user chose "Allow only once" option? And how could I request for "location always" permission? 

The flow below works fine if the user chose "location when in use", it just didn't work if the user chose "allow only once", many thanks!
```dart
    final locationWhenInUseStatus = await Permission.locationWhenInUse.request();
    print("Location when in use status: $locationWhenInUseStatus");
    if (locationWhenInUseStatus.isGranted) {
      print("Location when in use is granted");
      var locationAlwaysStatus = await Permission.locationAlways.request();
      print("After requesting location always");

      if (!locationAlwaysStatus.isGranted) {
        print("Location always is denied");
        bool res = await openAppSettings();
      }
      
    }
    if (!locationWhenInUseStatus.isGranted) {
      print("Location when in use is not granted");
    }
```

### LINKED ISSUE #1011: locationAlways permission isGranted returns true when location permission is set to "while using"
## 🐛 Bug Report

First off, thank you for taking the time to support this plugin. It's been a huge help for us and I know handling permissions on multiple platforms and versions is a pain.

We've been having some trouble around the locationAlways permission, and for us we're stuck on this problem on iOS: If the app's location permission is set to "while using", the result of `await Permission.locationAlways.isGranted` will return true. 

### Expected behavior
The expected behavior is that if the location permission is set to anything other than "always" (including "while using"), then `await Permission.locationAlways.isGranted` should return false.

### Reproduction steps
1. Request location in your app and select the "while using" option
2. Run the following statement after this `final isGranted = await Permission.locationAlways.isGranted`
3. Confirm that `isGranted` has a value of true

### Configuration
* iOS 15
* iPhone Mini 12
* Flutter 3.7.7
* Dart 2.19.1

**Version:** permission_handler - 10.2.0 

**Platform:** 
- [x] :iphone: iOS
- [ ] :robot: Android

### LINKED ISSUE #1018: locationAlways hangs if location was "Allowed Once" in iOS
## 🐛 Bug Report

Upon requesting `locationWhenInUse`, if the user presses "Allow Once" in iOS, the permission status returns `isGranted`, which, depending on how you interpret, might be a valid response.

However, this becomes a problem when you request `locationAlways` thereafter. If you do request `locationAlways` at this state, the request will hang.

If you go to the Settings and see the real permission status for the app, it'll still show "Ask Next Time Or When I Share", instead of "While Using the App". So from this point of view, I think `locationWhenInUse.isGranted` returning true might be an oversimplification and we might need another state to handle "Allow Once"

### Configuration

**Version:** 10.2.0

**Platform:** 
- [x] :iphone: iOS
- [ ] :robot: Android

mvanbeusekom:
The application will hang on the `Permission.locationAlways` future when requesting location always permissions directly after the user has granted "Allow once" permission.

This PR will fix this issue and makes sure the `PermissionStatus.denied` status is returned in these cases.

- Fixes #802
- Fixes #1011
- Fixes #1018

## Pre-launch Checklist

- [x] I rebased onto `main`.

### DIFF
--- a/permission_handler_apple/example/ios/Runner/Info.plist
+++ b/permission_handler_apple/example/ios/Runner/Info.plist
@@ -99,5 +99,9 @@
     <!-- Permission options for the `appTrackingTransparency` -->
     <key>NSUserTrackingUsageDescription</key>
     <string>appTrackingTransparency</string>
+	<key>CADisableMinimumFrameDurationOnPhone</key>
+	<true/>
+	<key>UIApplicationSupportsIndirectInputEvents</key>
+	<true/>
 </dict>
 </plist>
--- a/permission_handler_apple/example/ios/Runner/Info.plist
+++ b/permission_handler_apple/example/ios/Runner/Info.plist
@@ -99,9 +99,9 @@
     <!-- Permission options for the `appTrackingTransparency` -->
     <key>NSUserTrackingUsageDescription</key>
     <string>appTrackingTransparency</string>
-	<key>CADisableMinimumFrameDurationOnPhone</key>
-	<true/>
-	<key>UIApplicationSupportsIndirectInputEvents</key>
-	<true/>
+    <key>CADisableMinimumFrameDurationOnPhone</key>
+    <true/>
+    <key>UIApplicationSupportsIndirectInputEvents</key>
+    <true/>
 </dict>
 </plist>
--- a/permission_handler_apple/CHANGELOG.md
+++ b/permission_handler_apple/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 9.1.2
+
+* Fixes an issue where the `Permission.locationAlways.request()` call hangs when the application was granted "Allow once" permissions for fetching location coordinates.
+
 ## 9.1.1
 
 * Adds the new Android 13 permission "BODY_SENSORS_BACKGROUND" to PermissionHandlerEnums.h.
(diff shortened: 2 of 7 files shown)


permission_handler_apple/example/ios/Runner/Info.plist
```diff
@@ -99,5 +99,9 @@
     <!-- Permission options for the `appTrackingTransparency` -->
     <key>NSUserTrackingUsageDescription</key>
     <string>appTrackingTransparency</string>
+	<key>CADisableMinimumFrameDurationOnPhone</key>
```
JeroenWeener, line 102: Nit: is this supposed to be indented?

JeroenWeener (approved):

permission_handler_apple/example/ios/Runner/Info.plist
mvanbeusekom: No it is not, Xcode added it automatically but unfortunately used tabs where as the rest of the document uses spaces. Should be fixed now.

Issue #773, opened 2022-01-21
Check Bluetooth is Enable


themmfa:
## 💬 Questions and Help

For location, I can check if its enabled or disabled. However, I couldn't see a serviceStatus property for bluetooth. Is there any way to check if bluetooth is enabled or disabled?

wujek-srujek:
Would be nice to get this merged. In the meantime, I import `permission_handler_platform_interface: ^3.11.1` and use `PermissionHandlerPlatform.instance.checkServiceStatus(Permission.bluetooth)`.

JeroenWeener:
Hi @wujek-srujek, thanks for chiming in.

We are aiming to get this in soon.
See #1099, #1100, #1101, #1102 for progress on the feature.

Please note that the changes there exceed the changes in #942, as we found that the iOS implementation does not correctly return the service status. You may therefore find that you get `ServiceStatus.disabled`, regardless of the Bluetooth state, when using #942 in combination with `permission_handler_platform_interface: ^3.11.1` as you described.

JeroenWeener:
This issue is waiting for #1100 to be merged.

JeroenWeener:
Hi everyone!

The feature is now available on the latest version of the plugin. To pull it in, just run `flutter pub upgrade` in your project that has a dependency on either `permission_handler` or `permission_handler_apple`. Check the `pubspec.lock` file to make sure the dependency to `permission_handler_apple` is `9.1.4`. As `permission_handler_apple` is endorsed by `permission_handler`, you do not need to explicitly list it in your `pubspec.yaml` file.

PR #1100, opened 2023-07-11
Allow querying Bluetooth service status on iOS (Apple)

JeroenWeener:
This PR updates the Apple implementation, fixing an issue where the Bluetooth service status was never resolved properly. To this end it
- Makes `checkServiceStatus` async
- Keeps track of `PermissionStrategy` in `PermissionManager` (hold the reference to guarantee that the callback fires)

Additionally, it updates the displayed permission in the example app, as these were out of sync with the newest permissions.

This PR is part of feature #773.

## Pre-launch Checklist

- [x] I rebased onto `main`.

### DIFF
--- a/permission_handler_apple/ios/Classes/strategies/BluetoothPermissionStrategy.m
+++ b/permission_handler_apple/ios/Classes/strategies/BluetoothPermissionStrategy.m
@@ -11,8 +11,9 @@
 
 @implementation BluetoothPermissionStrategy {
     CBCentralManager *_centralManager;
-    PermissionStatusHandler _permissionStatusHandler;
     PermissionGroup _requestedPermission;
+    PermissionStatusHandler _permissionStatusHandler;
+    ServiceStatusHandler _serviceStatusHandler;
 }
 
 - (void)initManagerIfNeeded {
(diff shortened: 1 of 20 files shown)


permission_handler_apple/ios/Classes/strategies/BluetoothPermissionStrategy.m
```diff
@@ -23,23 +24,28 @@ - (void)initManagerIfNeeded {
...
     [self initManagerIfNeeded];
+    
+    _serviceStatusHandler = completionHandler;
+    _requestedPermission = permission;
+}
+
+- (void)handleCheckServiceStatusCallback {
     if (@available(iOS 10, *)) {
-        return [_centralManager state] == CBManagerStatePoweredOn ? ServiceStatusEnabled : ServiceStatusDisabled;
+        ServiceStatus serviceStatus = [_centralManager state] == CBManagerStatePoweredOn ? ServiceStatusEnabled : ServiceStatusDisabled;
```
mvanbeusekom, line 43: I am wondering if this really covers all situations. If the `[_centralManager state]` returns `[CBManagerStateUnauthorized]` we could argue that the service is enabled but the application is not authorised. Unfortunately the [documentation](https://developer.apple.com/documentation/corebluetooth/cbmanagerstate?language=objc) is not really clear on this. 

Did you by change test this situation specifically?

mvanbeusekom (changes requested):
The update to use a `completion` block looks really useful. The current implementation however does contain a bug (see comment).

Additionally I feel the `BluetoothPermissionStrategy.checkServiceStatus` might not cover all cases where the service is enabled.

permission_handler_apple/ios/Classes/strategies/BluetoothPermissionStrategy.m
JeroenWeener: I also cannot find any documentation on `CBManagerStateUnauthorized`. Currently, I'm receiving `CBManagerStatePoweredOn` and `CBManagerStatePoweredOff`, when Bluetooth is enabled/disabled, respectively, regardless of Bluetooth permission.

We could opt for considering `CBManagerStateUnauthorized` as `ServiceStatusEnabled`. What do you think?

JeroenWeener: I have tested the behavior. It works as follows:

- If the application has Bluetooth permission, everything works. We receive either `CBManagerStatePoweredOn` or `CBManagerStatePoweredOff`, depending on the Bluetooth service status.
- If the application was explicitly not granted access to Bluetooth (permission denied), we receive `CBManagerStateUnauthorized`, regardless of the Bluetooth service status.
- If the application has not requested Bluetooth permission yet, and therefore the permission state is unkown, the application will ask for permission first, and then carry out the logic as described in the two points above.

It seems that the last point is only true since iOS 13. https://support.apple.com/en-us/HT210578: 'Starting in iOS 13, iPadOS 13, watchOS 6, and tvOS 13, an app must ask permission to use Bluetooth functions'.

So basically, we can only reliably determine the Bluetooth status if permission is granted. Additionally, even checking the Bluetooth service status will prompt the user for permission.

What do you think, @mvanbeusekom? Would this be okay for the feature, provided we document this behavior?

mvanbeusekom: Agreed, if we document this clearly this should be fine.

mvanbeusekom (approved):
LGTM

PR #1099, opened 2023-07-11
Allow querying Bluetooth service status on iOS (platform interface)

JeroenWeener:
This PR updates the platform interface. It changes `Permission.bluetooth` from type `Permission` to type `PermissionWithService`, allowing iOS to provide the service's status.

Additionally, it cleans up the documentation of the permissions.

This PR is part of feature #773.

## Pre-launch Checklist

- [x] I made sure the project builds.
- [x] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [x] I updated `pubspec.yaml` with an appropriate new version according to the [pub versioning philosophy], or this PR is does not need version changes.
- [x] I updated `CHANGELOG.md` to add a description of the change.
- [x] I updated/added relevant documentation (doc comments with `///`).
- [x] I rebased onto `main`.
- [ ] I added new tests to check the change I am making, or this PR does not need tests.
- [x] I made sure all existing and new tests are passing.
- [x] I ran `dart format .` and committed any changes.
- [x] I ran `flutter analyze` and fixed any errors.

[Contributor Guide]: https://github.com/Baseflow/flutter-permission-handler/blob/master/CONTRIBUTING.md
[pub versioning philosophy]: https://dart.dev/tools/pub/versioning

PR #1102, opened 2023-07-11
Allow querying Bluetooth service status on iOS (main)

### LINKED ISSUE #773: Check Bluetooth is Enable

JeroenWeener:
This PR updates the dependencies to allow iOS to query the service status of the Bluetooth module.
Additionally, it updates the displayed permission in the example app, as these were out of sync with the newest permissions.

This PR depends on #1099, #1100, #1101, and closes #773.

## Pre-launch Checklist

- [x] I rebased onto `main`.


permission_handler/pubspec.yaml
```diff
@@ -22,10 +22,10 @@ dependencies:
   flutter:
     sdk: flutter
   meta: ^1.7.0
-  permission_handler_android: ^10.3.0
-  permission_handler_apple: ^9.1.1
+  permission_handler_android: ^10.3.1
+  permission_handler_apple: ^9.1.4
   permission_handler_windows: ^0.1.3
-  permission_handler_platform_interface: ^3.11.0
+  permission_handler_platform_interface: ^3.11.2
```
mvanbeusekom, line 28: This update is failing the build unnecessary. The updates are all patch updates which are already covered by the current version constraints. 

Currently the build fails because these versions do not exists (other PRs that are still in review). Rolling back the dependency updates will allow this PR to be moved forward without having to rely on the other PRs to be merged and published.

mvanbeusekom (commented):
Looks good but cannot be merged due to dependencies that have been updated. In general I think these updates are not necessary see also comment.

permission_handler/pubspec.yaml
JeroenWeener: @mvanbeusekom I updated the PR, removing the dependency updates. This means that only the example app is updated. Without explicitly stating the new versions of the `platform_interface` and `apple` packages, we should not note the change to the `Bluetooth` permission. I therefore updated the CHANGELOG as well.

How would we communicate with users that use `permission_handler` directly, that Bluetooth has been updated?

mvanbeusekom (approved):
LGTM

mvanbeusekom (approved):
LGTM

PR #1130, opened 2023-08-30
Return 'permanently denied' as permission status

### LINKED ISSUE #936: Permission.microphone.status returns denied when it should be permanentlyDenied
## 🐛 Bug Report

only on Android
the user permanentlyDenied the permission but the `get status` is returning denied.

### Expected behavior
the `get status` should return permanentlyDenied.

### Reproduction steps

deny the microphone permission twice. 
run `Permission.microphone.status` and see that the status is denied
run `Permission.microphone.request()` and see that the status is permanentlyDenied

### Configuration
-

**Version:** 9.2.0

**Platform:** 
- [ ] :iphone: iOS
- [X] :robot: Android

JeroenWeener:
Currently there is a discrepancy between a permission's status and its request result. Permission statuses are always 'denied', even if the request result is 'permanently denied'. This PR fixes that bug by leveraging the logic added in #1125.

@mvanbeusekom do you agree that a permission's status should be able to return 'permanently denied'? I noticed statuses can also return 'restricted', so maybe you want to treat a permission's status and its request result differently? On iOS, the status always returns 'permanently denied' (note that permissions can only be asked once on iOS).
@mvanbeusekom should this change be labeled as breaking? Some folks might be depending on the exact return value of a permission's status.

This PR fixes #936. Note that the mentioned issue does not seem to be particular to only the microphone permission, but rather, all permissions that show the user a dialog (so not the ones that open the settings page).

## Pre-launch Checklist

- [x] I rebased onto `main`.

### DIFF
--- a/permission_handler_android/android/src/main/java/com/baseflow/permissionhandler/PermissionManager.java
+++ b/permission_handler_android/android/src/main/java/com/baseflow/permissionhandler/PermissionManager.java
@@ -42,7 +42,7 @@ final class PermissionManager implements PluginRegistry.ActivityResultListener, 
     /**
      * The number of pending permission requests.
      * <p>
-     * This number is set by {@link this#requestPermissions(List, Activity, RequestPermissionsSuccessCallback, ErrorCallback)}
+     * This number is set by {@link this#requestPermissions(List, RequestPermissionsSuccessCallback, ErrorCallback)}
      * and then reduced when receiving results in {@link this#onActivityResult(int, int, Intent)}
      * and {@link this#onRequestPermissionsResult(int, String[], int[])}.
      */
@@ -54,7 +54,7 @@ final class PermissionManager implements PluginRegistry.ActivityResultListener, 
      * {@link this#onActivityResult(int, int, Intent)} and
      * {@link this#onRequestPermissionsResult(int, String[], int[])}.
      * It is (re)initialized when new permissions are requested through
-     * {@link this#requestPermissions(List, Activity, RequestPermissionsSuccessCallback, ErrorCallback)}.
+     * {@link this#requestPermissions(List, RequestPermissionsSuccessCallback, ErrorCallback)}.
      */
     private Map<Integer, Integer> requestResults;
 
@@ -284,13 +284,11 @@ final class PermissionManager implements PluginRegistry.ActivityResultListener, 
      * requested through this method were handled, and if so, return the result back to Dart.
      *
      * @param permissions     the permissions that are requested.
-     * @param activity        the activity.
      * @param successCallback the callback for returning the permission results.
      * @param errorCallback   the callback to call in case of an error.
      */
     void requestPermissions(
         List<Integer> permissions,
-        Activity activity,
         RequestPermissionsSuccessCallback successCallback,
         ErrorCallback errorCallback) {
         if (pendingRequestCount > 0) {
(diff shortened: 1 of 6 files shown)


permission_handler_android/android/src/main/java/com/baseflow/permissionhandler/PermissionManager.java
```diff
@@ -306,25 +311,25 @@ void requestPermissions(
     @PermissionConstants.PermissionStatus
     private int determinePermissionStatus(
             @PermissionConstants.PermissionGroup int permission,
-            Context context) {
+            Activity activity) {
```
mvanbeusekom, line 314: This change might result into issue because the `activity` might be `null` in certain conditions (e.g. when the app is running in the background). 

I noticed this method uses the activity a few times without making sure it is not `null` which will result into runtime exceptions. I think we should provide an alternative solution, maybe using casting where we validate if the supplied `context` is of type `Activity` and if so do more advanced status validation, if not stick to the current implementation.

mvanbeusekom (changes requested):
Hi @JeroenWeener,

Thanks for this PR. If we can be sure the returning the `permanentlyDenied` status works reliable than that would be a great addition and would help a lot of users. So yes, this would be great. This should however be considered a breaking change. Like you mentioned we should assume there are apps that apply won't handle `permanentlyDenied` status on Android.

I have also added one concern about updating the `Context context` parameter of the `checkPermissionStatus` method, please have a look at this.

mvanbeusekom (approved):
LGTM, however lets first merge in pr #1139

jjoelson:
Hi @JeroenWeener and @mvanbeusekom!

Here is a useful writeup from the `geolocator` package's wiki that explains why they don't feel they can reliably distinguish between denied and permanently denied: https://github.com/Baseflow/flutter-geolocator/wiki/Breaking-changes-in-7.0.0#android-permission-update

The full writeup is worth reading, but the long and short of it is that since the "Ask every time" option was added in Android API 30, the user manually selecting that option in settings will break the denied vs permanently denied logic.

The `permission_handler` package itself referenced that write up in its [6.0.0 change notes](https://github.com/Baseflow/flutter-permission-handler/wiki/Changes-in-6.0.0) and says the following: 
> In the old implementation permission_handler saved the status in Android's Shared Preferences to determine if the permission was permanently denied. The implementation now does not save this info, so this needs to be checked twice, to determine permanently denied permissions.

Has something changed since then to enable this package to reliably distinguish denied vs permanently denied, or is the implementation in this PR vulnerable to the same bug described in the above writeup?

Tomorrow or next week I can do some tests with the latest `permission_handler` to see if any bugs crop up relating to this change, but I just wanted to put this on your radar.

JeroenWeener:
Hi @jjoelson thank you for bringing this up!

I was unaware of this history of the `geolocator` when I worked on this issue. What I get from the wiki is that the solution is slightly different. I have experimentally tested the changes made in this PR back then. A short write-up is available in the code: https://github.com/Baseflow/flutter-permission-handler/blob/384fc9ea32d2db0ba84d2983ab0bfe62ec850e19/permission_handler_android/android/src/main/java/com/baseflow/permissionhandler/PermissionUtils.java#L399-L458

I am very interested in your test results to see if you can verify correct behavior 🙌

jjoelson:
Hi @JeroenWeener! I did some testing and unfortunately I did encounter the bug. I filed a new issue here with repro steps: https://github.com/Baseflow/flutter-permission-handler/issues/1206

maquejp:
Hello, 
The bug is still there... Any idea when you can work on it?

 checkPermission: LocationPermission.denied requestPermission: LocationPermission.deniedForever
 
Thank you.

mvanbeusekom:
Hello @maquejp, 

As mentioned by Jeroen Weener this is not a bug, rather a limitation of Android. You can also read our reasoning to this in our Geolocator post: https://github.com/Baseflow/flutter-geolocator/wiki/Breaking-changes-in-7.0.0#android-permission-update (also mentioned in the comment from jjoelson.

Issue #938, opened 2022-10-13
nearbyWifiDevices.request() does not work


guyluz11:
## 🐛 Bug Report

Request nearby WiFi devices do not open any popup for me and returned denied value.

### Expected behavior
Popup to allow access opens up or function returns automatically granted state.

### Reproduction steps
1. Create empty project
2. Add a call to the function `await Permission.nearbyWifiDevices.request();`
3. Add a bunch of android permissions to `AndroidManifest.xml` just to make sure that we are not missing anything 
```
    <uses-permission android:name="android.permission.INTERNET"/>
    <uses-permission android:name="android.permission.ACCESS_NETWORK_STATE" />
    <uses-permission android:name="android.permission.CHANGE_NETWORK_STATE" />
    <uses-permission android:name="android.permission.ACCESS_WIFI_STATE" />
    <uses-permission android:name="android.permission.CHANGE_WIFI_STATE" />
    <uses-permission android:name="android.permission.NEARBY_WIFI_DEVICES" />
    <uses-permission android:name="android.permission.WRITE_SETTINGS" />
    <uses-permission android:name="android.permission.ACCESS_FINE_LOCATION" />
    <uses-permission android:name="android.permission.ACCESS_COARSE_LOCATION" />
```
4. Run the app on android 12 device

### Configuration

**Version:** 10.1.0

**Platform:** 
- [ ] :iphone: iOS
- [x] :robot: Android

guyluz11:
Ok so no nearby wifi devices permission is needed for below 13.

I do expect the method to return true in that case and for this to be stated in the documentation of the method (not just inside code 31).
![image](https://user-images.githubusercontent.com/9304740/196750821-a9936e49-2440-4336-9566-1e9df8a5752b.png)

JeroenWeener:
This seems to be resolved in the latest version of the plugin: https://github.com/Baseflow/flutter-permission-handler/blob/main/permission_handler_platform_interface/lib/src/permissions.dart#L256C1-L257C53.

```dart
  /// Permission for connecting to nearby devices via Wi-Fi (Android T+ only).
  static const nearbyWifiDevices = Permission._(31);
```

guyluz11:
What is Android `Android T+`? there aren't any relevant results when searching it in google.
Why not write an API-level or android version instead?

JeroenWeener:
[...]

PR #1131, opened 2023-08-31
Updates Android version mentions

JeroenWeener:
Android and iOS version constraints can be a tricky thing to take into account when developing using the `permission handler` plugin. It should be clear at a glance to developers which permissions can be used with which versions of Android or iOS. To that end, this PR updates the mentions of Android versions to follow a strict convention that should speak most to the mind of developers. The convention is 'Android $name (API $version)'. For example: 'Android 13 (API 33)' or 'Android Marshmallow (API 23)'. Both the public name and the API version are common ways of referring to Android versions. By listing them both, we cater to developers who are familiar with the names as they are presented to Android users, and to developers who are familiar with the API versions. Additionally, by mentioning them both, we avoid confusion, as later Android versions use a number as name. For example 'Android 10' vs 'API 10' (API 29 and Android Gingerbread, respectively). For an overview of Android's version history, see https://en.wikipedia.org/wiki/Android_version_history.

One instance of a user being confused by the current Android version naming: https://github.com/Baseflow/flutter-permission-handler/issues/938#issuecomment-1699424150

## Pre-launch Checklist

- [x] I rebased onto `main`.

### DIFF
--- a/permission_handler_android/CHANGELOG.md
+++ b/permission_handler_android/CHANGELOG.md
@@ -1,3 +1,8 @@
+## 10.3.4
+
+* Fixes a bug where the permission status would return 'permanently denied'
+instead of 'denied' when the user would dismiss the permission dialog.
+
 ## 10.3.3
 
 * Migrates the Gradle compile arguments to the example app, so they are not enforced upon consumers of the plugin.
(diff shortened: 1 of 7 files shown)


JeroenWeener:
Closing this PR as it has been split in #1135 and #1136.

PR #1139, opened 2023-09-01
Handle special permissions correctly when requesting multiple permissions

JeroenWeener:
# Context
Android distinguishes between [runtime permissions](https://developer.android.com/guide/topics/permissions/overview#runtime) and [special permissions](https://developer.android.com/guide/topics/permissions/overview#special). Runtime permissions give an app additional access to restricted data or let the app perform restricted actions that more substantially affect the system and other apps. These permissions present the user with a dialog where they can choose to grant or deny the permission. Special permissions guard access to system resources that are particularly sensitive or not directly related to user privacy. These permissions are requested by sending an `Intent` to the OS. The OS will open a special settings page where the user can grant the permission. Runtime permission request results will be reported through `onRequestPermissionsResult(int, String[], int[])`, while special permissions request results will be reported through `onActivityResult(int, int, Intent)`.

# Problem statement
Currently, the code submits all permission request results for runtime permissions at the same time, and the results of special permissions separately. This is a mistake, as we can only return the results to Dart once. Instead, we should keep track of the number of pending requests and only return the results once all of them are in. This PR implements exactly that.

Closes #539, #761, #1126

## Pre-launch Checklist

- [x] I made sure the project builds.
- [x] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [x] I updated `pubspec.yaml` with an appropriate new version according to the [pub versioning philosophy], or this PR is does not need version changes.
- [x] I updated `CHANGELOG.md` to add a description of the change.
- [x] I updated/added relevant documentation (doc comments with `///`).
- [x] I rebased onto `main`.
- [ ] I added new tests to check the change I am making, or this PR does not need tests.
- [x] I made sure all existing and new tests are passing.
- [x] I ran `dart format .` and committed any changes.
- [x] I ran `flutter analyze` and fixed any errors.

[Contributor Guide]: https://github.com/Baseflow/flutter-permission-handler/blob/master/CONTRIBUTING.md
[pub versioning philosophy]: https://dart.dev/tools/pub/versioning

Issue #1126, opened 2023-08-18
[Reply already submitted][Bug]: Android crash : Reply already submitted


Passer-by:
### Please check the following before submitting a new issue.

- [x] I have searched the [existing issues](https://github.com/baseflow/flutter-permission-handler/issues).
- [X] I have carefully [read the documentation](https://github.com/Baseflow/flutter-permission-handler/blob/main/permission_handler/README.md) and verified I have added the required platform specific configuration.

### Please select affected platform(s)

- [X] Android
- [ ] iOS
- [ ] Windows

### Steps to reproduce

Run permission_handler/example
Simultaneously tap ignoreBatteryOptimizations and notification.
Perform any action

### Expected results

App crash

### Actual results

The app is running normally.

### Code sample

<details><summary>Code sample</summary>

```dart
[Paste your code here]
```

</details>

### Screenshots or video

<details>
<summary>Screenshots or video demonstration</summary>

https://github.com/Baseflow/flutter-permission-handler/assets/58846244/1a8c66e2-3b19-4e33-a5d5-40f4ba7da284

</details>

### Version

master 10.4.3

### Flutter Doctor output

<details><summary>Doctor output</summary>

```console
[√] Flutter (Channel stable, 3.10.6, on Microsoft Windows [版本 10.0.19044.3324], locale zh-CN)
    • Flutter version 3.10.6 on channel stable at D:\code\flutter\SDK\flutter_windows_3.10.0-stable\flutter
    • Upstream repository https://github.com/flutter/flutter.git
    • Framework revision f468f3366c (5 weeks ago), 2023-07-12 15:19:05 -0700
    • Engine revision cdbeda788a
    • Dart version 3.0.6
    • DevTools version 2.23.1
    • Pub download mirror https://pub.flutter-io.cn

[√] Windows Version (Installed version of Windows is version 10 or higher)

[√] Android toolchain - develop for Android devices (Android SDK version 33.0.2)
    • Android SDK at D:\code\android\AndroidSDK
    • Platform android-33-ext4, build-tools 33.0.2
    • Java binary at: D:\code\android\AndroidStudio\jbr\bin\java
    • Java version OpenJDK Runtime Environment (build 17.0.6+0-b2043.56-9586694)
    • All Android licenses accepted.

[√] Chrome - develop for the web
    • Chrome at C:\Program Files\Google\Chrome\Application\chrome.exe

[...]

PR #1147, opened 2023-09-07
Reply already submitted in Android 13 while user request permissions

jinchengwu-ola:
Reply already submitted in Android 13 while user request permissions (eg. post_notification, locations)

*List at least one fixed issue.*
Fatal Exception: java.lang.RuntimeException
Failure delivering result ResultInfo{who=@android:requestPermissions:, request=24, result=-1, data=Intent { act=android.content.pm.action.REQUEST_PERMISSIONS (has extras) }} to activity: java.lang.IllegalStateException: Reply already submitted

io.flutter.embedding.engine.dart.DartMessenger$Reply.reply (DartMessenger.java:435)
io.flutter.plugin.common.MethodChannel$IncomingMethodCallHandler$1.success (MethodChannel.java:263)
com.baseflow.permissionhandler.g.a (Unknown Source:2)
com.baseflow.permissionhandler.n.onRequestPermissionsResult (PermissionManager.java:173)
io.flutter.embedding.engine.FlutterEngineConnectionRegistry$FlutterEngineActivityPluginBinding.onRequestPermissionsResult (FlutterEngineConnectionRegistry.java:779)
io.flutter.embedding.engine.FlutterEngineConnectionRegistry.onRequestPermissionsResult (FlutterEngineConnectionRegistry.java:411)
io.flutter.embedding.android.FlutterActivityAndFragmentDelegate.onRequestPermissionsResult (FlutterActivityAndFragmentDelegate.java:761)
io.flutter.embedding.android.FlutterActivity.onRequestPermissionsResult (FlutterActivity.java:795)

## Pre-launch Checklist

- [ ] I updated/added relevant documentation (doc comments with `///`).
- [x] I rebased onto `main`.
- [ ] I made sure all existing and new tests are passing.
- [ ] I ran `flutter analyze` and fixed any errors.

### DIFF
--- a/permission_handler_android/android/src/main/java/com/baseflow/permissionhandler/PermissionManager.java
+++ b/permission_handler_android/android/src/main/java/com/baseflow/permissionhandler/PermissionManager.java
@@ -162,7 +162,7 @@ final class PermissionManager implements PluginRegistry.ActivityResultListener, 
         }
 
         if (permissions.length == 0 && grantResults.length == 0) {
-            Log.w(PermissionConstants.LOG_TAG, "onRequestPermissionsResult is called without results");
+            Log.w(PermissionConstants.LOG_TAG, "onRequestPermissionsResult is called without results. This is probably caused by interfering request codes. If you see this error, please file an issue in flutter-permission-handler, including a list of plugins used by this application: https://github.com/Baseflow/flutter-permission-handler/issues");
             return false;
         }
 
--- a/permission_handler_android/android/src/main/java/com/baseflow/permissionhandler/PermissionManager.java
+++ b/permission_handler_android/android/src/main/java/com/baseflow/permissionhandler/PermissionManager.java
@@ -161,6 +161,11 @@ final class PermissionManager implements PluginRegistry.ActivityResultListener, 
             return false;
         }
 
+        if (permissions.length == 0 && grantResults.length == 0) {
+            Log.w(PermissionConstants.LOG_TAG, "onRequestPermissionsResult is called without results");
+            return false;
+        }
+
         for (int i = 0; i < permissions.length; i++) {
             final String permissionName = permissions[i];
 
--- a/permission_handler_android/CHANGELOG.md
+++ b/permission_handler_android/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 11.0.1
+
+* Fixes `java.lang.IllegalStateException: Reply already submitted` when requesting post notification permission.
+
 ## 11.0.0
 
 * **BREAKING CHANGE:** Fixes a bug where the permission status would return 'denied' regardless of whether the status was 'denied' or 'permanently denied'.
--- a/permission_handler_android/pubspec.yaml
+++ b/permission_handler_android/pubspec.yaml
@@ -1,7 +1,7 @@
 name: permission_handler_android
 description: Permission plugin for Flutter. This plugin provides the Android API to request and check permissions.
 homepage: https://github.com/baseflow/flutter-permission-handler
-version: 11.0.0
+version: 11.0.1
 
 environment:
   sdk: ">=2.15.0 <4.0.0"


JeroenWeener:
Hi @jinchengwu-ola, thank you for contributing to the `flutter-permission-handler` plugin!

It is not clear to me which bug your changes are supposed to fix. Could you link us to the issue number?
We recently merged #1139, which might solve the issue you are targeting. Could you confirm this?

jinchengwu-ola:
> Hi @jinchengwu-ola, thank you for contributing to the `flutter-permission-handler` plugin!
> 
> It is not clear to me which bug your changes are supposed to fix. Could you link us to the issue number? We recently merged #1139, which might solve the issue you are targeting. Could you confirm this?

Hi @JeroenWeener , I encounter an issue requesting post_notification permission which will cause 'Reply already submitted' cashing in Android 13, Galaxy S21, xiaomi devices.  

It crashes at com.baseflow.permissionhandler.n.onRequestPermissionsResult (PermissionManager.java:173). 
And I found that this system recall with empty list params (param permissions is empty, param grantResults is empty, normally it shouldn't be empty,  and I ignored this recall than it would not crash).  

I guess it is an system pitfall that special permission requesting's result will not only be recall in onActivityResult, but also in onRequestPermissionsResult at the same time. But still need to verify to check Android 13 system source code. 

Anyway It looks not exactly  like merged #1139  mentioned. Crashes at which  merged #1139 mentioned happen in onActivityResult(PermissionManager.java:64). 

After all, I found @Shilk1266 have the similar issue with me, bellow is the link.
https://github.com/Baseflow/flutter-permission-handler/issues/1126#issuecomment-1704534923

Hope I make it clear, thanks~

JeroenWeener:
Hi @jinchengwu-ola thanks for your reply!

This is an interesting issue. We'd be interested in including your PR in the code base.

Before we do that though, we would like to unravel what is causing this issue in the first place. Our suspicion is that maybe another plugin in your project is using the same identifier for requesting permissons. To that end, could you share your `pubspec.yaml` file? Or if you implemented any Android code in your project that would be interesting to look into as well.

permission_handler_android/android/src/main/java/com/baseflow/permissionhandler/PermissionManager.java
```diff
@@ -125,6 +125,10 @@ public boolean onRequestPermissionsResult(int requestCode, String[] permissions,
            return false;
         }
 
+        if (permissions.length ==0 && grantResults.length==0){
```
JeroenWeener, line 128:
```suggestion
        if (permissions.length == 0 && grantResults.length == 0) {
```

JeroenWeener, line 128: Could you add a `Log.w` call here that tells the developer that this callback was called without results?

Your changes make sure the application will not crash. Also, this case should never happen. But in case it does, like in the issue you reported in the comments, we would like the developer to be aware of it so they can raise an issue. Otherwise we risk that the plugin fails silently, which can be quite confusing when debugging.

JeroenWeener (changes requested):
Could you please update the version number in `pubspec.yaml` and add a changelog entry in `CHANGELOG.md`?

jinchengwu-ola:
> Could you please update the version number in `pubspec.yaml` and add a changelog entry in `CHANGELOG.md`?

the version of permissiopn_handler is [10.4.3]

permission_handler_android/android/src/main/java/com/baseflow/permissionhandler/PermissionManager.java
```diff
@@ -161,6 +161,11 @@ public boolean onRequestPermissionsResult(
             return false;
         }
 
+        if (permissions.length == 0 && grantResults.length == 0) {
+            Log.w(PermissionConstants.LOG_TAG, "onRequestPermissionsResult is called without results");
```
JeroenWeener, line 165:
```suggestion
            Log.w(PermissionConstants.LOG_TAG, "onRequestPermissionsResult is called without results. This is probably caused by interfering request codes. If you see this error, please file an issue in flutter-permission-handler, including a list of plugins used by this application: https://github.com/Baseflow/flutter-permission-handler/issues");
```

JeroenWeener (changes requested):

JeroenWeener (approved):
Thank you for your contribution @jinchengwu-ola! We hope to see more of you in the future 😄

PR #1148, opened 2023-09-07
Fix Unhandled Exception: type 'int' is not a subtype of type 'FutureO…

jinchengwu-ola:
*Exception throw while calling shouldShowRequestPermissionRationale in ios*

*List at least one fixed issue.*
dart Unhandled Exception: type 'int' is not a subtype of type 'FutureOr<bool>'

## Pre-launch Checklist

- [ ] I updated `pubspec.yaml` with an appropriate new version according to the [pub versioning philosophy], or this PR is does not need version changes.
- [x] I rebased onto `main`.
- [ ] I made sure all existing and new tests are passing.
- [ ] I ran `dart format .` and committed any changes.
- [ ] I ran `flutter analyze` and fixed any errors.

### DIFF
--- a/permission_handler_platform_interface/lib/src/method_channel/method_channel_permission_handler.dart
+++ b/permission_handler_platform_interface/lib/src/method_channel/method_channel_permission_handler.dart
@@ -90,7 +90,10 @@ class MethodChannelPermissionHandler extends PermissionHandlerPlatform {
       Permission permission) async {
     final shouldShowRationale = await _methodChannel.invokeMethod(
         'shouldShowRequestPermissionRationale', permission.value);
-
+    // calling this on ios will return an int value 0, this will cause type cast error
+    if (shouldShowRationale is int){
+      return shouldShowRationale > 0;
+    }
     return shouldShowRationale ?? false;
   }
 }


permission_handler_platform_interface/lib/src/method_channel/method_channel_permission_handler.dart
```diff
@@ -90,7 +90,10 @@ class MethodChannelPermissionHandler extends PermissionHandlerPlatform {
       Permission permission) async {
     final shouldShowRationale = await _methodChannel.invokeMethod(
         'shouldShowRequestPermissionRationale', permission.value);
-
+    // calling this on ios will return an int value 0, this will cause type cast error
+    if (shouldShowRationale is int){
+      return shouldShowRationale > 0;
+    }
```
mvanbeusekom, line 96: Missed this completely, however since this is very Android specific implementation it would be better to add the following at the top of the method:

```dart
    if (defaultTargetPlatform != TargetPlaform.android) {
        return false;
    }
```
This would also solve the issue and on top prevent method calls to other platforms that might not be implemented and throw an unimplemented error.

mvanbeusekom (changes requested):
Thank you for contributing and fixing this issue, I have completely missed it. Now that you have pointed it out, I think there might be a better way to solve this issue as per my suggestion below.

Also would you please update the `CHANGELOG.md` and `pubspec.yaml` files as mentioned in the PR checklist?

mvanbeusekom:
Superseded by #1298

Issue #719, opened 2021-11-03
[Android] permission request may return wrong result


Perfetto2020:
## 🐛 Bug Report

Permission request may got wrong result on Android platform.

Let's say two permissions declared in AndroidManifest.xml of a flutter app:
``` xml
    <uses-permission android:name="android.permission.READ_PHONE_STATE" />
    <uses-permission android:name="android.permission.READ_PHONE_NUMBERS" />
```
Both of them are `Permission.phone` on permission_handler dart side.

On devices running Android 11 and after (Samsung Galaxy Note 20 on my hand)，grant `READ_PHONE_STATE` but deny `READ_PHONE_NUMBERS` to the app, then when you request `Permission.phone` permission on app dart side by `[Permission.phone].request()`,  you got the result `{Permission.phone: PermissionStatus.granted}` which is incorrect or inaccurate.

### Expected behavior
Give caller precise status of every manifest permissions, or
give denied status if any of them are denied.

### Reproduction steps
as described above

### Configuration

**Version:** 8.2.5

**Platform:** 
- [ ] :iphone: iOS *not test*
- [x] :robot: Android

Perfetto2020:
The codes produce this issue is in `onRequestPermissionsResult` of `PermissionManager.java`
``` java
            } else if (!requestResults.containsKey(permission)) {
                requestResults.put(
                        permission,
                        PermissionUtils.toPermissionStatus(this.activity, permissionName, result));
           }
```
Only the status of the first manifest permission of a `PermissionConstants.PermissionGroup` will be recorded in the `requestResults` to represent the overall status of the `PermissionConstants.PermissionGroup`.
It's an oversight or just by design?

sashker:
This problem is very frustrating. Does anybody have a patch or a workaround?

PR #1157, opened 2023-09-19
Enhancement/split bind call redirection service permission

### LINKED ISSUE #360: await Permission.phone.request().isGranted always be true
## 🐛 Bug Report

```dart
var status = await Permission.phone.request();
print(status); // status always be granted

if (await Permission.phone.request().isGranted) {
  // always be true
}
```

**Version:** 5.0.1+1

**Platform:** 
- [ ] :iphone: iOS
- [x] :robot: Android

JeroenWeener:
Requesting the permission status of `Phone` always returns `denied`, because we are checking for `BIND_CALL_REDIRECTION_SERVICE`. This is a system app permission and should not be checked. This PR removes this permission from the manifest and the business logic. When implementing, I found another bug, likely #719, where permissions in Flutter that require multiple permissions in Android would return based off only one permission result that was returned. I made some changes that checks all results and selects the most restrictive status. For example, when a permission in Flutter requires 4 permissions in Android, we might receive `GRANTED`, `NEVER_ASK_AGAIN`, `DENIED` and `GRANTED`. The result will then be `NEVER_ASK_AGAIN`.

Closes #360, #719.

## Pre-launch Checklist

- [x] I rebased onto `main`.

### DIFF
--- a/permission_handler_android/CHANGELOG.md
+++ b/permission_handler_android/CHANGELOG.md
@@ -1,3 +1,8 @@
+## 11.0.2
+
+* Fixes a bug where `Permission.Phone` would always return 'denied' when requesting the permission status.
+* Fixes a bug where Flutter permissions that require multiple Android permissions would base their status on the status of the first Android permission, as opposed to the result of all relevant Android permissions.
+
 ## 11.0.1
 
 * Fixes `java.lang.IllegalStateException: Reply already submitted` when requesting post notification permission.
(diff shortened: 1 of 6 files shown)


permission_handler_android/CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
+## 12.0.0
+
+* **BREAKING CHANGES:**
+  * Fixes a bug where `Permission.Phone` would always return 'denied' when requesting the permission status.
+  * Fixes a bug where Flutter permissions that require multiple Android permissions would base their status on the status of the first Android permission, as opposed to the result of all relevant Android permissions.
```
mvanbeusekom, line 5: I am wondering if these are really breaking changes. They seem like bugs where users already have different expectations from how the current version works and these fixes align to the expectations. 

Personally I think we can make this a patch release.

mvanbeusekom (changes requested):
LGTM, however I feel this can be a patch release.

mvanbeusekom (approved):
LGTM

PR #1161, opened 2023-09-22
Update documentation in permission_status.dart

bramp:
Minor edit to update the documentation to be clearer when limited and provisional statuses may be returned. The docs are repeated ~3 times, so I made them consistent.

## Pre-launch Checklist

- [x] I rebased onto `main`.


permission_handler_platform_interface/lib/src/permission_status.dart
```diff
@@ -12,19 +12,34 @@ enum PermissionStatus {
...
+  /// Android 11+ (API 30+): whether the user denied the permission for a second
+  /// time.
+  /// Below Android 11 (API 30): whether the user denied access to the requested
+  /// feature and selected to never again show a request.
+  ///
+  /// *On iOS:*
+  /// If the user has denied access to the requested feature.
+  ///
+  /// WARNING: This can only be determined AFTER requesting this permission.
+  /// Therefore make a `request` call first.
```
mvanbeusekom, line 39: Since version 11.0.0 this is no longer relevant, we have added some logic that would cache the permanentlyDenied permission and will also return it when calling `checkPermission`.

permission_handler_platform_interface/lib/src/permission_status.dart
```diff
@@ -119,25 +143,33 @@ extension FuturePermissionStatusGetters on Future<PermissionStatus> {
...
   /// Below Android 11 (API 30): whether the user denied access to the requested
   /// feature and selected to never again show a request.
-  /// The user may still change the permission status in the settings.
   ///
   /// *On iOS:*
   /// If the user has denied access to the requested feature.
-  /// The user may still change the permission status in the settings
+  ///
+  /// WARNING: This can only be determined AFTER requesting this permission.
+  /// Therefore make a `request` call first.
```
mvanbeusekom, line 160: Same as before, since version 11.0.0 this is no longer the case.

mvanbeusekom (changes requested):
Hi @bramp,

Thank you very much for this pull-request, the documentation changes are great. Just one small thing regarding the permanentlyDenied status which has been update since version 11.0.0 of the plugin.

Also would you please update the CHANGELOG and pubspec.yaml with a patch version update?

permission_handler_platform_interface/lib/src/permission_status.dart
bramp: ack.

permission_handler_platform_interface/lib/src/permission_status.dart
bramp: ack.

bramp:
Thanks for the review.

Fixed the docs since v11.0.0, and  bumped the version, and added a entry to the CHANGELOG. I originally didn't bump the version, as the documentation change didn't seem worthy of a new version.

bramp:
oh wait, I just stumbled across @template and @macro [dartdoc annotations](https://pub.dev/packages/dartdoc#macros), that would enable me to remove all the duplication, e.g 

```dart
enum PermissionStatus {
  /// {@template denied}
  /// The user denied access to the requested feature, permission needs to be
  /// asked first.
  /// {@endtemplate}
  denied,
```

```dart
extension PermissionStatusGetters on PermissionStatus {
  /// {@macro denied}
  bool get isDenied => this == PermissionStatus.denied;
```

That generates correct dartdocs. However, looking in my IDE (vscode) it does not process the macro, and instead shows nothing.

Do you have a preference if we should move to the template/macro? or just live with a little bit of copy and paste :)

mvanbeusekom (approved):
LGTM, thank you very much for taking the effort and making this contribution

PR #1164, opened 2023-09-23
Refactor Permission Handling for Improved Readability

michaelnabil230:
This PR introduces a new style for checking permission status.

Before:

```dart
Future<bool> askPermission() async {
  PermissionStatus status = await Permission.contacts.request();
  if (status.isDenied) {
    // Your code
  } else if (status.isGranted) {
    // Your code
  } else if (status.isRestricted) {
    // Your code
  }
}
```

After:

```dart
Future<bool> askPermission() async {
  await Permission.contacts
    .onDeniedCallback(() {
      // Your code
    })
    .onGrantedCallback(() {
      // Your code
    })
    .onRestrictedCallback(() {
      // Your code
    })
    .request();
}
```

In this updated code, the permission status checks have been replaced with a more concise style that leverages the `PermissionCallbacks` extension. This change simplifies the code and makes it more readable.

## Pre-launch Checklist

- [X] I made sure the project builds.
- [X] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [X] I updated/added relevant documentation (doc comments with `///`).
- [X] I rebased onto `main`.
- [X] I added new tests to check the change I am making, or this PR does not need tests.
- [X] I made sure all existing and new tests are passing.
- [X] I ran `dart format .` and committed any changes.
- [X] I ran `flutter analyze` and fixed any errors.

### DIFF
--- a/permission_handler/lib/permission_handler.dart
+++ b/permission_handler/lib/permission_handler.dart
@@ -1,3 +1,5 @@
+import 'dart:async';
+
 import 'package:flutter/foundation.dart';
 import 'package:permission_handler_platform_interface/permission_handler_platform_interface.dart';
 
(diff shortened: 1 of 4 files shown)


mvanbeusekom (changes requested):
Hi @michaelnabil230,

This is a really interesting addition that we would love to support. I have two requests that I hope you could help update:

1. The different callbacks currently accept a function of type `VoidCallback`, I can imagine that some developers would like to run asynchronous tasks in side the callback (e.g. fetching a location once location permission is granted). I think it would be beneficial to update these callbacks to support a `Future<void>` (or possible a `FutureOr<void>`) method.
2. With the current changes there are two methods to start the permission request, `request` and `ask`. I think this might lead to confusion as it will be possible to register callbacks and then use `request` which doesn't use the registered callbacks. Since this is not a breaking change, why not add the callbacks to the current `PermissionAction` extension and call them from the `request` method?

michaelnabil230:
Hi @mvanbeusekom

Thank you for your instruction. 
Now, you have the opportunity to merge the PR so that everyone can reap the benefits of this feature.

michaelnabil230:
Hi @mvanbeusekom

Thank you for your instruction.
Now, you have the opportunity to merge the PR so that everyone can reap the benefits of this feature.

permission_handler/lib/permission_handler.dart
```diff
@@ -21,6 +23,60 @@ Future<bool> openAppSettings() => _handler.openAppSettings();
 
 /// Actions that can be executed on a permission.
 extension PermissionActions on Permission {
+  /// Callback for when permission is denied.
+  static FutureOr<void>? Function()? onDenied;
```
mvanbeusekom, line 27: I think we should make these fields private (by starting their names with an `_`), as they are accessed only from internal implementation and should not be exposed outside of the extension.

mvanbeusekom (changes requested):
Hi @michaelnabil230,

I have noticed one more issue that I feel should be addressed.

permission_handler/lib/permission_handler.dart
michaelnabil230: Sounds like a good idea.

michaelnabil230:
Hi @mvanbeusekom 

Thanks for the update! It's awesome to hear that the code is ready to merge. Let's proceed with the merging process.

mvanbeusekom (approved):
LGTM, thank you for this contribution.

P.S. maybe you could create another PR updating the documentation and explain this alternative route, so people are informed about this new feature.

Issue #631, opened 2021-07-23
Add support for Linux


guyluz11:
Currently this package dose not support Linux.

Is there any intention to add support for it?.

![image](https://user-images.githubusercontent.com/9304740/126724232-32a5d395-b800-4d36-8be5-576bf77e6b3d.png)

SoYoureAWaffleMan:
I agree it's a bit frustrating. This error doesn't explain much:

```
MissingPluginException(No implementation found for method checkPermissionStatus on channel flutter.baseflow.com/permissions/methods)
```

If Linux is not supported at all then an Exception stating "Linux is not supported" would at least be clear.

In my first usage I'm happy for the bool `Permission.notification.isGranted` to default to true if the platform is not supported, so providing a simple way to do that and documenting it would be sufficient. As it is I'll have to catch an obscure exception and set the default there.

JeroenWeener:
We would consider adding support for Linux, although it is not our primary concern at this moment. If someone experienced with the Linux OS wants to take on this challenge, feel free to submit a PR. The Baseflow team can offer assistance where needed.

PR #1175, opened 2023-09-28
Support a permissive default implementation of permission_handler for linux.

patlefort:
* Assume that all permissions are granted.
* Assume that all services are enabled.
* Assume that there is no "app settings" and permission requests.
* Make it the default for linux platform.

This will enable support for the linux platform. Linux has plenty of ways to sandbox an application but no singular way. Later I could look into flatpak and snaps and see if it can be integrated. For now this plugin should at least make apps work for linux. I named it the "default" plugin but I can always rename it if preferred.

It should fix issue #631

## Pre-launch Checklist

- [x] I rebased onto `main`.

### DIFF
--- a/permission_handler/CHANGELOG.md
+++ b/permission_handler/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 11.0.1
+
+* Add linux platform support with `permission_handler_default`.
+
 ## 11.0.0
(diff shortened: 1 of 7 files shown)


EduMenges:
Any updates on this?

PR #1180, opened 2023-10-06
Pigeon + method invocation example

JeroenWeener:
Adds the Pigeon setup for the refactor.

To obtain an `Activity` instance, one can call `Android.onActivityChanged((Activity? activity) => ...)`. With this activity, native Android code can be called. As an example, `ActivityCompat.shouldShowRequestPermissionRationale(Activity activity, String permissionName)` is implemented and can be called with the obtained activity.

## Pre-launch Checklist

- [ ] I updated `pubspec.yaml` with an appropriate new version according to the [pub versioning philosophy], or this PR is does not need version changes.
- [ ] I rebased onto `main`.


permission_handler_android/example/lib/main.dart
```diff
@@ -8,6 +9,9 @@ void main() {
       githubURL: 'https://github.com/Baseflow/flutter-permission-handler',
       pubDevURL: 'https://pub.dev/packages/permission_handler',
       pages: [PermissionHandlerWidget.createPage()]));
+
+  // TODO(jweener): what is the proper way to register a platform implementation?
+  PermissionHandlerAndroid.registerWith();
```
mvanbeusekom, line 14: The proper way to do so is by updating the `pubspec.yaml` of the permission_handler_android package. The `dartPluginClass` field informs the Flutter engine on which class it should call the `registerWith` method.

For example the snippet below informs the Flutter engine to call `registerWith` on the `PermissionHandlerAndroid` Dart class:

```yaml
flutter:
  plugin:
    implements: permission_handler
    platforms:
      android:
        package: com.baseflow.permissionhandler
        pluginClass: PermissionHandlerPlugin
        dartPluginClass: PermissionHandlerAndroid
```

mvanbeusekom (changes requested):
Hi @JeroenWeener,

The structure looks really good, exactly how I envisioned it. I made one remark on the `onDeactivateActivity` method which should remove the activity from the instance manager. 

I also explained the "proper" way on how to register the correct implementation class.

permission_handler_android/lib/src/android_permission_handler.dart
```diff
@@ -0,0 +1,74 @@
...
+    ActivityFlutterApi.setup(api);
+  }
+
+  /// Registers this class as the default instance of [PermissionHandlerPlatform].
+  static void registerWith() {
+    final PermissionHandlerAndroid permissionHandlerAndroid =
+        PermissionHandlerAndroid();
+    permissionHandlerAndroid.setUp();
+    PermissionHandlerPlatform.instance = permissionHandlerAndroid;
+  }
```
mvanbeusekom, line 44:
```suggestion
  factory PermissionHandlerAndroid() {
        final PermissionHandlerAndroid instance = PermissionHandlerAndroid();
        instance.setUp();
        return instance;
  }

  static void registerWith() {
    PermissionHandlerPlatform.setOnCreatePlatformInstance(
      builder: () => PermissionHandlerAndroid(),
    );
  }
```

permission_handler_android/lib/src/android.dart
```diff
@@ -0,0 +1,56 @@
...
+  }) {
+    // Instantiate Activity Flutter API only once.
+    _activityFlutterApi ??= activityFlutterApi;
+    if (_activityFlutterApi == null) {
+      _activityFlutterApi = ActivityFlutterApiImpl();
+      ActivityFlutterApi.setup(_activityFlutterApi);
+    }
+
+    _activityFlutterApi!.registerOnActivityChanged(onActivityChanged);
+  }
```
mvanbeusekom, line 55: Wouldn't it be clearer to split this method up in a `onAttachedToActivity` and `onDetachedFromActivity`? 

Also I think we could add a `currentActivity` field or property which returns the current activity if one is attached. This would allow developers to get the current activity whenever they need it and don't have to rely on the `onAttachedToActivity` and `onDetachedFromActivity` methods.

This means we also don't have to immediately call the callbacks when they are registered.

permission_handler_android/lib/src/android_permission_handler_api_impls.dart
```diff
@@ -0,0 +1,103 @@
...
+  /// Registers a callback that is invoked with the current activity, everytime
+  /// it changes.
+  void registerOnActivityChanged(
+    void Function(Activity? activity) onActivityChanged,
+  ) {
+    // Emit the current attached activity.
+    onActivityChanged(_activity);
+
+    _onActivityChangedCallbacks.add(onActivityChanged);
+  }
```
mvanbeusekom, line 71: If we support registering multiple callbacks (which I think we should), we should also support removing a callback when users no longer want to listen. 

I think we should also rename the method to `addOnAttachedToActivityCallback` and create a matching method `removeOnAttachedToActivityCallback`.

mvanbeusekom (changes requested):

permission_handler_android/android/src/main/java/com/baseflow/permissionhandler/PermissionHandlerPlugin.java
```diff
@@ -1,85 +1,61 @@
...
     @Override
     public void onAttachedToActivity(@NonNull ActivityPluginBinding binding) {
-        startListeningToActivity(
-            binding.getActivity()
-        );
-
-        this.pluginBinding = binding;
-        registerListeners();
+        this.activity = binding.getActivity();
+        activityFlutterApi.create(this.activity);
```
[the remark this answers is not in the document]
mvanbeusekom: As discussed off-line the `ActivityFlutterApiImpl` has no direct connection with one specific `Activity` therefore my suggestion is not valid and should not be applied.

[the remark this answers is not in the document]
JeroenWeener: The `ActivityFlutterApiImpl` class is responsible for creating and disposing `Activity`s, but it is not tied to a specific `Activity`. The `PermissionHandlerPlugin` class **is** associated with a single `Activity`, namely, the one that attaches to the Flutter engine. Therefore, I believe it makes more sense to keep track of the activity in the `PermissionHandlerPlugin` as opposed to the `ActivityFlutterApiImpl`.

permission_handler_android/lib/src/android.dart
JeroenWeener: I am afraid that providing direct access to the `currentActivity` will be a source for bugs in consuming applications. Developers might not appreciate that the activity might become stale. With the current implementation, it is clear to the developers that the `Activity` is volatile, and they should program with this in mind.

For example: `Android.onActivityChanged` can be called by the developer directly, but is also called from `PermissionHandlerAndroid`. It can be unclear for developers whether the activity has attached already. For example, when depending solely on `permission_handler_android`, the developer needs to call `onAttachedToActivity`, as the suggested `currentActivity` will be `null`. If depending on `permission_handler` instead, `onAttachedToActivity` is already called by `PermissionHandlerAndroid`, and `currentActivity` **might** not be `null`, depending on when they fetch the activity. Concluding, it can be very opaque for developers how they should get ahold of an activity.

The way I see it, with the suggested approach, the developer would be tempted to write logic to tackle this uncertainty like this:

```dart
final Activity? activity = Android.currentActivity;
if (activity != null) {
  ...
} else {
  Android.onAttachedToActivity((Activity activity) => ...);
}
```

permission_handler_android/lib/src/android_permission_handler_api_impls.dart
```diff
@@ -0,0 +1,135 @@
...
+  final List<void Function()> _onDetachedFromActivityCallbacks = [];
+
+  /// Adds a callback to be called when an activity is attached.
+  ///
+  /// If an activity is attached when this method is called, the callback is
+  /// called immediately.
+  void addOnAttachedToActivityCallback(
+    void Function(Activity attachedActivity) onActivityAttached,
+  ) {
+    if (_activity != null) onActivityAttached(_activity!);
```
mvanbeusekom, line 69: According to the Google Flutter convention `if` statements should always have braces, the one exception is when you directly return.

```suggestion
    if (_activity != null) { 
      onActivityAttached(_activity!);
    }
```

mvanbeusekom (changes requested):
Hi Jeroen,

Two small remarks but looks good otherwise.

mvanbeusekom (approved):
LGTM

PR #1184, opened 2023-10-11
Correct the grammar in the documents and restructure the formatting of all the markdowns

michaelnabil230:
Before marge, please review the corrected files


mvanbeusekom (changes requested):
Hi @michaelnabil230,

Thank you very much for the thorough  walkthrough and corrections. One thing that I would like to mention is that we use `*` to indicate items in a bulleted list, especially in the CHANGELOG.md files (so it complies with the [convention Google](https://github.com/flutter/flutter/wiki/Contributing-to-Plugins-and-Packages#changelog-style) is using for Flutter). It would be great if you could update the change to reflect this.

michaelnabil230:
Hello, @mvanbeusekom
It appears that you should adhere to the writing guidelines outlined in the Markdown guide rather than Google's specific writing rules. Each company typically has its own distinct writing style for its documents. We should follow the approach established by the Markdown guide when collaborating with them, but when not actively engaged with these companies, we are expected to utilize the writing style recommended by the Markdown guide.

Reference: https://www.markdownguide.org/basic-syntax/

Screenshot 2023-10-21 at 5 13 18 PM

michaelnabil230:
Hi @mvanbeusekom, 

I am not convinced about this change, but as you see. I converted all `-` to the previous

michaelnabil230:
Hi @mvanbeusekom, 

I am not convinced about this change, but as you see. I converted all `-` to the previous.

Can merge the PR if all changes are correct

mvanbeusekom (approved):
LGTM, thank you for this contribution and applying the requested changes.

Issue #1191, opened 2023-10-14
[Question]: Request "ignoreBatteryOptimizations" and allow, but return denied.


StarAndSun:
### Please check the following before submitting a new issue.

- [X] I have searched the [existing issues](https://github.com/baseflow/flutter-permission-handler/issues).
- [X] I have carefully [read the documentation](https://github.com/baseflow/flutter-permission-handler/blob/main/permission_handler/README.md).

### Please select for which platform(s) you need help

- [X] Android
- [ ] iOS
- [ ] Windows

### Your question

I use android 13 on my oppo phone. In the Manifest by adding < USES - permission android: name = "android. Permission. REQUEST_IGNORE_BATTERY_OPTIMIZATIONS" / >. And use PermissionStatus state = await Permission. IgnoreBatteryOptimizations. Request (); . In the popup, Allow is selected, but denied is returned. Hoping to get some advice.

### Version

11.0.1

mvanbeusekom:
Hi @StarAndSun,

Thank you for reporting this issue. I was able to reproduce the issue on my Google Pixel 7a using the permission_handler example app. 

This is something we will look into.

Coimbra1984:
I have the same problem on a Samsung A53. My workaround is calling Permission.ignoreBatteryOptimizations.request().isGranted twice.

MSOB7YY:
just for additional info, it prints `D/permissions_handler( 7414): No permissions found in manifest for: []16` in the terminal

mvanbeusekom:
@StarAndSun, @Coimbra1984, this issue is resolved in version 12.0.1 of the permission_handler_android plugin.

anandrex5:
For me it works -

Add this dependency to the pubspec.yaml - disable_battery_optimization: ^1.1.1

and when you press any Button that request battery saver settings -

ElevatedButton(
  onPressed: () async {
    await Utils(context).requestBatteryOptimization();
  },
  child: Text('Request Battery Optimization'),
)


and this is the Util method, which opens the battery-saver settings where you can select No restrictions  -

import 'package:disable_battery_optimization/disable_battery_optimization.dart';


// Request battery optimization settings
  Future<void> requestBatteryOptimization() async {
    try {
      bool isBatteryOptimizationDisabled =
          await DisableBatteryOptimization.isBatteryOptimizationDisabled ?? false;

      if (!isBatteryOptimizationDisabled) {
        // Open settings to disable battery optimization
[...]

PR #1200, opened 2023-10-23
Report ignoreBatteryOptimizations permission correctly

mvanbeusekom:
Fixes a bug where the `ignoreBatteryOptimizations` reports a denied status when requesting permission and the user granted permission.

Solves #1191

## Pre-launch Checklist

- [x] I rebased onto `main`.

### DIFF
--- a/permission_handler_android/android/src/main/java/com/baseflow/permissionhandler/PermissionManager.java
+++ b/permission_handler_android/android/src/main/java/com/baseflow/permissionhandler/PermissionManager.java
@@ -35,15 +35,12 @@ import io.flutter.plugin.common.PluginRegistry;
 
 final class PermissionManager implements PluginRegistry.ActivityResultListener, PluginRegistry.RequestPermissionsResultListener {
 
-    @Nullable
-    private RequestPermissionsSuccessCallback successCallback;
-
-    @Nullable
-    private Activity activity;
-
     @NonNull
     private final Context context;
-
+    @Nullable
+    private RequestPermissionsSuccessCallback successCallback;
+    @Nullable
+    private Activity activity;
     /**
      * The number of pending permission requests.
      * <p>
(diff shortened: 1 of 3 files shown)


permission_handler_android/android/src/main/java/com/baseflow/permissionhandler/PermissionManager.java
```diff
@@ -77,14 +74,20 @@ public boolean onActivityResult(int requestCode, int resultCode, Intent data) {
...
-            : PermissionConstants.PERMISSION_STATUS_DENIED;
-
-        int permission;
+        int status, permission;
 
         if (requestCode == PermissionConstants.PERMISSION_CODE_IGNORE_BATTERY_OPTIMIZATIONS) {
             permission = PermissionConstants.PERMISSION_GROUP_IGNORE_BATTERY_OPTIMIZATIONS;
+
+            if (Build.VERSION.SDK_INT < Build.VERSION_CODES.M) {
+                return false;
```
JeroenWeener, line 83: Do we want to report `false` or `true` when the device runs a version below Marshmallow? Doze and App Stand-by were not introduced until Marshmallow. It might make sense to return this as `true` from a developer's perspective, as to indicate that the application is not subjected to battery optimizations.

JeroenWeener (changes requested):
The code looks good and appears to be working.
I have one comment regarding the workings of the permission when running Android below Marshmallow.

permission_handler_android/android/src/main/java/com/baseflow/permissionhandler/PermissionManager.java
mvanbeusekom: That is a good point, I will update the code

JeroenWeener: This is probably also the case for the logic that handles just checking the status of the permission

JeroenWeener (approved):

Issue #1239, opened 2023-11-27
[Bug]: permission_handler_html does not work in a non-https web app


ekuleshov:
### Please check the following before submitting a new issue.

- [X] I have searched the [existing issues](https://github.com/baseflow/flutter-permission-handler/issues).
- [X] I have carefully [read the documentation](https://github.com/Baseflow/flutter-permission-handler/blob/main/permission_handler/README.md) and verified I have added the required platform specific configuration.

### Please select affected platform(s)

- [ ] Android
- [ ] iOS
- [ ] Windows

### Steps to reproduce

Checkout the `flutter-permission-handler` project from github.

Navigate to `permission_handler_html/example` and run the following command:

```
flutter run -d chrome --web-hostname=<host ip address>
```

where `<host ip address>` is the non-localhost of the machine, e.g. a public DHCP address like 192.168.1.x.

This is essentially as if the web app is running from a non-https location and for this command the example app is failing to start printing the following error to the console:

```
Rejecting promise with error: Unexpected null value.
```

I don't know any workarounds for this and it looks like this issue is in the `permission_handler_html` plugin registration code.

### Expected results

The `flutter-permission-handler` plugin should work in dev/localhost, https and non-https web apps.

### Actual results

The example web app fail to launch.

### Code sample

Checkout the `flutter-permission-handler`. Navigate to `permission_handler_html/example` and run the following command:

```
flutter run -d chrome --web-hostname=<host ip address>
```

where `<host ip address>` is the non-localhost of the machine, e.g. a public DHCP address like 192.168.1.x.

### Version

0.1.0+1

### Flutter Doctor output

<details><summary>Doctor output</summary>

```console
flutter doctor --verbose
[✓] Flutter (Channel stable, 3.16.0, ***)
    • Flutter version 3.16.0 on channel stable at /***/flutter
    • Upstream repository https://github.com/flutter/flutter.git
    • Framework revision db7ef5bf9f (12 days ago), 2023-11-15 11:25:44 -0800
    • Engine revision 74d16627b9
    • Dart version 3.2.0
    • DevTools version 2.28.2

[✓] Chrome - develop for the web
    • Chrome at /Applications/Google Chrome.app/Contents/MacOS/Google Chrome

[✓] VS Code (version 1.84.2)
    • VS Code at /Applications/Visual Studio Code.app/Contents
    • Flutter extension can be installed from:
[...]

PR #1240, opened 2023-11-27
Html non secure permissions

ekuleshov:
This should fix https://github.com/Baseflow/flutter-permission-handler/issues/1239

## Pre-launch Checklist

- [ ] I updated/added relevant documentation (doc comments with `///`).
- [x] I rebased onto `main`.

### DIFF
--- a/permission_handler_html/lib/web_delegate.dart
+++ b/permission_handler_html/lib/web_delegate.dart
@@ -64,8 +64,12 @@ class WebDelegate {
   }
 
   Future<bool> _requestMicrophonePermission() async {
+    if (_devices == null) {
+      return false;
+    }
+
     try {
-      html.MediaStream? mediaStream = await _devices?.getUserMedia({'audio': true});
+      html.MediaStream mediaStream = await _devices!.getUserMedia({'audio': true});
 
       // In browsers, calling [getUserMedia] will start the recording
       // automatically right after. This is undesired behavior as
@@ -74,8 +78,8 @@ class WebDelegate {
       // The manual stop action is then needed here for to stop the automatic
       // recording.
 
-      if (mediaStream?.active ?? false) {
-        final audioTracks = mediaStream!.getAudioTracks();
+      if (mediaStream.active ?? false) {
+        final audioTracks = mediaStream.getAudioTracks();
         if (audioTracks.isNotEmpty) {
           audioTracks[0].stop();
         }
@@ -88,8 +92,12 @@ class WebDelegate {
   }
 
   Future<bool> _requestCameraPermission() async {
+    if (_devices == null) {
+      return false;
+    }
+
     try {
-      html.MediaStream? mediaStream = await _devices?.getUserMedia({'video': true});
+      html.MediaStream? mediaStream = await _devices!.getUserMedia({'video': true});
 
       // In browsers, calling [getUserMedia] will start the recording
       // automatically right after. This is undesired behavior as
@@ -98,8 +106,8 @@ class WebDelegate {
       // The manual stop action is then needed here for to stop the automatic
       // recording.
 
-      if (mediaStream?.active ?? false) {
-        final videoTracks = mediaStream!.getVideoTracks();
+      if (mediaStream.active ?? false) {
+        final videoTracks = mediaStream.getVideoTracks();
         if (videoTracks.isNotEmpty) {
           videoTracks[0].stop();
         }
--- a/permission_handler_html/lib/web_delegate.dart
+++ b/permission_handler_html/lib/web_delegate.dart
@@ -97,7 +97,7 @@ class WebDelegate {
     }
 
     try {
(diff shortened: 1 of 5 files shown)


TimHoogstrate:
Dear @ekuleshov,

I saw your issue and the solution seems to be fine. However, I am not able to test the PR at the moment and I lack the required knowledge about the topic. Could you explain why this is happening and why it is now resolved?

Kind regards,

TimHoogstrate (commented):
Code LGTM

ekuleshov:
> I saw your issue and the solution seems to be fine. However, I am not able to test the PR at the moment and I lack the required knowledge about the topic. Could you explain why this is happening and why it is now resolved?

This issue is actually pretty bad... As soon as you add the current version of `flutter-permission-handler` to the project yaml or even add some other plugin that has `flutter-permission-handler` as a transitive dependency, the web app fail to start when launched from a non-https location (in our case it was an internal web app and https is hard to setup).

From HTML docs for the [mediaDevices](https://developer.mozilla.org/en-US/docs/Web/API/Navigator/mediaDevices) property:

> Secure context: This feature is available only in [secure contexts](https://developer.mozilla.org/en-US/docs/Web/Security/Secure_Contexts) (HTTPS), in some or all [supporting browsers](https://developer.mozilla.org/en-US/docs/Web/API/Navigator/mediaDevices#browser_compatibility).

For some reason the `WebPermissionHandler` is accessing the `html.window.navigator.mediaDevices` in the static field initializer and forcing value as non-null. That is where it throws a null pointer exception, which is not easy to see, unless you launch a non obfuscated web app in debugger with `--web-hostname=<host ip address>` option using non-localhost ip address.

mvanbeusekom:
Hi @ekuleshov,

Thank you for this PR, upon request from @TimHoogstrate I did another code review. 

First I think you are correct that the static `_devices` field should be optional as per your explanation and supplied fixes. This however will not completely resolve the issue as the problem will now be moved down the line into the `WebDelegate` class. An `Unexpected null value` exception will now be thrown when someone is trying to request camera or microphone permissions as they rely on the `devices` variable not to be `null` (which is a design issue that should be fixed).

It would be great if you could update this PR and have the `_requestCameraPermission` and `_requestMicrophonePermission` methods accept an nullable instance of `html.MediaDevices` as a parameter and have them `return false;` when the `devices` parameter is `null`. 

If you don't have the time please let us know, in that case we would be happy to take over the PR and add the necessary changes.

permission_handler_html/lib/web_delegate.dart
```diff
@@ -77,8 +74,8 @@ class WebDelegate {
       // The manual stop action is then needed here for to stop the automatic
       // recording.
 
-      if (mediaStream.active!) {
-        final audioTracks = mediaStream.getAudioTracks();
+      if (mediaStream?.active ?? false) {
+        final audioTracks = mediaStream!.getAudioTracks();
         if (audioTracks.isNotEmpty) {
           audioTracks[0].stop();
         }
```
mvanbeusekom, line 81: Combining the condition here mixes two different situations and treating them the same, which leads to unexpected behaviour.

1. The original test is just to make sure the `mediaStream` is stopped if it is started (which is caused by calling the `getUserMedia()` method on line 68).
2. The added tests checks if `mediaStream` is `null`, which is most likely the result of `_devices` being `null`. 

I feel we should `return false;` at the start of the method if `_devices` is `null` and prevent this mixed behaviour. Added benefit will be that returning `false` is translated to a `PermissionStatus.denied` which is was it comes down to in a HTTP (non-secure) context.

permission_handler_html/lib/web_delegate.dart
```diff
@@ -103,8 +98,8 @@ class WebDelegate {
       // The manual stop action is then needed here for to stop the automatic
       // recording.
 
-      if (mediaStream.active!) {
-        final videoTracks = mediaStream.getVideoTracks();
+      if (mediaStream?.active ?? false) {
```
mvanbeusekom, line 101: As with the microphone permissions, we should `return false;` earlier in my opinion

mvanbeusekom (changes requested):

permission_handler_html/lib/web_delegate.dart
ekuleshov: > 2\. The added tests checks if `mediaStream` is `null`, which is most likely the result of `_devices` being `null`

Not only. The `mediaStream.active` value is also nullable. Though a good catch on `PermissionStatus.denied`. 

The original behaviour to return `true` by default is now how I'd do it, but I didn't want to completely rewrite this. Perhaps someone could improve it when doing refactoring to remove dependency on `html` packages in order to support wasm...

mvanbeusekom (approved):
LGTM!

Thank you for the additional testing and of course contributing with these fixes. It is really appreciated.

Issue #1280, opened 2024-02-09
[Question]: Does permission_handler need to implement Apple's Required Reason API or Privacy Manifest support?


ShuheiSuzuki-07:
### Please check the following before submitting a new issue.

- [X] I have searched the [existing issues](https://github.com/baseflow/flutter-permission-handler/issues).
- [X] I have carefully [read the documentation](https://github.com/baseflow/flutter-permission-handler/blob/main/permission_handler/README.md).

### Please select for which platform(s) you need help

- [ ] Android
- [X] iOS
- [ ] Windows

### Your question

Does permission_handler need to implement Apple's Required Reason API or Privacy Manifest support?

https://github.com/flutter/flutter/issues/131940#issuecomment-1749876448

### Version

any

TimHoogstrate:
Related to (https://github.com/Baseflow/flutter-geolocator/issues/1360)

ChristianEdwardPadilla:
FYI Apple has indicated an enforcement date of May 1st for Privacy Manifests.

mvanbeusekom:
Version 9.4.1 of the permission_handler_apple package was just released which includes the privacy manifest. For more details see #1291.

stuartmorgan:
Per https://github.com/Baseflow/flutter-permission-handler/issues/1292 the current manifest is not complete.

PR #1291, opened 2024-03-18
Adds iOS privacy manifest.

mvanbeusekom:
Adds the required privacy manifest to the iOS implementation.

Because the permission_handler_apple package doesn't collect any privacy data, the privacy manifest can be empty. 

Note that we are following the standard Flutter / Cocoapods implementation which has the following caveat:

> The large caveat is that we do not currently know if this actually works. This is the method of inclusion that seems to be [the consensus among people using Cocoapods](https://github.com/CocoaPods/CocoaPods/issues/10325), as bundling it directly as a `resource` causes problems for clients who do not use `use_frameworks`. (In theory it seems like a manifest would not actually be *required* in that case since there is no framework, but it has the potential to actually stomp top-level resources.) Hopefully the automated analysis that Apple will eventually roll out will tolerate the file being bundled in a resource bundle in the framework rather than a top-level manifest file. If not, however, it's not clear how Cocoapods can be supported, so we can adopt this common approach for now under the assumption that eventually tooling will adapt to the reality of the ecosystem, and revisit the exact bundling later if necessary. 

Source: https://github.com/flutter/packages/commit/c5349bc9a5be1cbef9c13571d9d6d2b506240b20

Solves #1280 

## Pre-launch Checklist

- [x] I rebased onto `main`.

### DIFF
--- a/permission_handler_apple/CHANGELOG.md
+++ b/permission_handler_apple/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 9.4.1
+
+* Adds privacy manifest.
+
 ## 9.4.0
 
 * Adds a new permission `Permission.backgroundRefresh` to check the background refresh permission status.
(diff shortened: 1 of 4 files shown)


stuartmorgan:
> Because the permission_handler_apple package doesn't collect any privacy data, the privacy manifest can be empty.

This is not correct; see https://github.com/Baseflow/flutter-permission-handler/issues/1292

PR #1316, opened 2024-05-01
Remove references to v1 embedding

johnmccutchan:
Flutter will be removing the remains of the android v1 embedding code and to make that easier for you here is a CL. 

Please make a release ASAP that includes this change.

### DIFF
--- a/permission_handler_android/android/src/main/java/com/baseflow/permissionhandler/PermissionHandlerPlugin.java
+++ b/permission_handler_android/android/src/main/java/com/baseflow/permissionhandler/PermissionHandlerPlugin.java
@@ -30,30 +30,6 @@ public final class PermissionHandlerPlugin implements FlutterPlugin, ActivityAwa
     @Nullable
     private MethodCallHandlerImpl methodCallHandler;
 
-    /**
-     * Registers a plugin implementation that uses the stable {@code io.flutter.plugin.common}
-     * package.
-     *
-     * <p>Calling this automatically initializes the plugin. However plugins initialized this way
-     * won't react to changes in activity or context, unlike {@link PermissionHandlerPlugin}.
-     */
-    @SuppressWarnings("deprecation")
-    public static void registerWith(io.flutter.plugin.common.PluginRegistry.Registrar registrar) {
-        final PermissionHandlerPlugin plugin = new PermissionHandlerPlugin();
-
-        plugin.pluginRegistrar = registrar;
-        plugin.permissionManager = new PermissionManager(registrar.context());
-        plugin.registerListeners();
-
-        plugin.startListening(registrar.context(), registrar.messenger());
-
-        if (registrar.activeContext() instanceof Activity) {
-            plugin.startListeningToActivity(
-                registrar.activity()
-            );
-        }
-    }
-
     @Override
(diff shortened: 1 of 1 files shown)


johnmccutchan:
@mvanbeusekom PTAL

johnmccutchan:
@ChristianEdwardPadilla PTAL

johnmccutchan:
This is the most up to date issue https://github.com/flutter/flutter/issues/143531

Please do not delay in landing this PR and making a new release of your plugin.

johnmccutchan:
To be clear, this code was deprecated six years ago and should have been removed from your codebase ~5 years ago

mvanbeusekom:
I disagree, it is marked deprecated in our code base as well. And there were still plenty of users running flutter pre-1.12 a few years ago.

The code still compiles and is not used if the app uses v2 of Android embedding.

Having a quick Look at the PR it is missing the required changes to the pubspec.yaml and CHANGELOG.md as mentioned in the PR template's checklist.

johnmccutchan:
We will be removing all support for this long deprecated API very soon. You can choose to keep this code in your plugin but it will break in the future.

mvanbeusekom:
Well your tone is kinda rude which doesn't work in your favor.

Also the PR is missing required changes to be eligible to merge and yet you are pushing us to merge which isn't helping.

johnmccutchan:
> Well your tone is kinda rude which doesn't work in your favor.
> Also the PR is missing required changes to be eligible to merge and yet you are pushing us to merge which isn't helping.

As a favor, I prepared this PR to make things easier for you. You stated that you disagreed with this change:

> I disagree, it is marked deprecated in our code base as well. And there were still plenty of users running flutter pre-1.12 a few years ago.
> The code still compiles and is not used if the app uses v2 of Android embedding.

I am informing you that if you don't remove this code from your plugin it will be broken in the near future.  Let me be clear: Apps using v2 of the Android embedding will be broken if they use your plugin.

gmackall:
Just to add some context here, as I'm the one primarily working on the removal of the v1 embedding. 

V1 apps will be fully broken in the upcoming stable release, due to https://github.com/flutter/flutter/pull/144726 (it will no longer be possible to build them with the Flutter tool). In the stable after that we plan to remove the v1 embedding completely, which means that any plugin which references the v1 embedding will fail to compile. 

We are recommending that all plugins remove all references to the v1 embedding ahead of that time to avoid the scramble of fixing the crash, making a new release, and then having users of the plugin upgrade to that new version. I'd be interested to hear if you have a reason not to adopt this change, or if there is more info I could add to help convince you. 

Let me know, I'm happy to help!

mvanbeusekom:
Hi @gmackall, 

Thank you for providing the context.

I am not against the change at all it is just that @johnmccutchan was pretty direct and without any context I wasn't confident the change is legit.

Anyway I will look into it first thing tomorrow. One question, should I mark this a breaking change?

gmackall:
If it lands before the next stable, it would be fair to describe it as a breaking change. 

It would also be fine to land concurrently with the next stable (or shortly thereafter), in which case I don't think you would need to mark it as such given that the only users you would be breaking would already be fully broken by the tool.

PR #1346, opened 2024-07-15
Android check request results null

shlowdy:
In some specific situations, requestResults will be null.


diegofrata:
This PR would fix #1360. 
@mvanbeusekom any chance this could be merged?

Issue #1386, opened 2024-09-24
[Bug]: web  DartError: TypeError: null: type 'Null' is not a subtype of type 'JSObject'


402162095:
### Please check the following before submitting a new issue.

- [ ] I have searched the [existing issues](https://github.com/baseflow/flutter-permission-handler/issues).
- [X] I have carefully [read the documentation](https://github.com/Baseflow/flutter-permission-handler/blob/main/permission_handler/README.md) and verified I have added the required platform specific configuration.

### Please select affected platform(s)

- [ ] Android
- [ ] iOS
- [ ] Windows

### Steps to reproduce

DartError: TypeError: null: type 'Null' is not a subtype of type 'JSObject'
dart-sdk/lib/_internal/js_dev_runtime/private/ddc_runtime/errors.dart 297:3       throw_
dart-sdk/lib/_internal/js_shared/lib/rti.dart 1385:3                              _failedAsCheck
dart-sdk/lib/_internal/js_shared/lib/rti.dart 1363:3                              _generalAsCheckImplementation
dart-sdk/lib/_internal/js_shared/lib/js_util_patch.dart 81:5                      getProperty
packages/permission_handler_html/permission_handler_html.dart 13:65               get _devices
dart-sdk/lib/_internal/js_dev_runtime/private/ddc_runtime/operations.dart 1293:8  get
packages/permission_handler_html/permission_handler_html.dart 24:9                registerWith
web_plugin_registrant.dart 50:24                                                  registerPlugins
web_entrypoint.dart 27:24                                                         <fn>
lib/ui_web/ui_web/initialization.dart 37:9                                        <fn>
dart-sdk/lib/_internal/js_dev_runtime/patch/async_patch.dart 84:54                runBody
dart-sdk/lib/_internal/js_dev_runtime/patch/async_patch.dart 127:5                _async
lib/ui_web/ui_web/initialization.dart 35:16                                       <fn>
dart-sdk/lib/_internal/js_dev_runtime/private/ddc_runtime/operations.dart 550:37  _checkAndCall
dart-sdk/lib/_internal/js_dev_runtime/private/ddc_runtime/operations.dart 620:10  callMethod
dart-sdk/lib/_internal/js_dev_runtime/private/ddc_runtime/operations.dart 624:5   dsend
lib/_engine/engine/app_bootstrap.dart 31:18                                       autoStart
[...]

PR #1407, opened 2024-11-18
Bug fix for the DartError: TypeError: null: type 'Null' is not a subtype of type 'JSObject'

dg76:
This PR fixes the bug mentioned here:

https://github.com/Baseflow/flutter-permission-handler/issues/1386

It seems that "mediaDevices" is only available when using https or localhost, but not when using e.g. http://0.0.0.0. Thus it is necessary to check if "mediaDevices" exists before accessing it.

### DIFF
--- a/permission_handler_html/CHANGELOG.md
+++ b/permission_handler_html/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 0.1.3+4
+
+- http compatibility: Don't crash on missing `window.navigator.mediaDevices` property
+
 ## 0.1.3+3
 
 - Safari < 16 compatibility: Don't crash on missing `window.navigator.permissions` property
--- a/permission_handler_html/lib/permission_handler_html.dart
+++ b/permission_handler_html/lib/permission_handler_html.dart
@@ -1,5 +1,6 @@
 import 'dart:async';
 import 'dart:js_interop_unsafe';
+import 'dart:js_util' as js_util;
 
 import 'package:web/web.dart' as web;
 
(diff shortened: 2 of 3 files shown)


mvanbeusekom:
Hi @dg76,

Thank you for this contribution, it makes a lot of sense and I can indeed reproduce the issue. The suggested update looks really good, however do you mind updating the version number (increasing the patch number by one) and add an entry to the CHANGELOG.md (according to the [changelog style guidelines](https://github.com/flutter/flutter/blob/master/docs/ecosystem/contributing/README.md#changelog-style))?

dg76:
No problem, I have changed the CHANGELOG.md and pubspec.yaml. It should appear as a second commit in the same pull request. Is it good now? Thank you for your quick response!

mvanbeusekom (approved):

mvanbeusekom:
Thanks again, I have just published version 0.1.3+4 of the permission_handler_html package to pub containing this fix.

dg76:
Thank you very much, I have configured my project to use 0.1.3+4 and it works correctly now. Thank you very much for publishing the new version so quickly!

LukasMirbt:
Hi! Unfortunately it seems like the `'dart:js_util` import breaks wasm compilation.

dg76:
I am sorry, I think I have fixed it with this PR: https://github.com/Baseflow/flutter-permission-handler/pull/1411

I am now using a method to access "mediaDevices" that was already used for "permissions" in the same file.

mvanbeusekom:
I have just published version 0.1.3+5 of the permission_handler_html package containing the fix provided by @dg76 

Thanks @dg76 for the quick fix.

Issue #1409, opened 2024-11-20
[permission_handler_html]: 0.1.3+4 breaks wasm compilation with `dart:js_util` import


LukasMirbt:
### Is there an existing issue for this?

- [X] I have searched the [existing issues](https://github.com/baseflow/flutter-permission-handler/issues).

### Old behavior

`flutter build web --wasm` worked

### Current behavior

`flutter build web --wasm` fails with:
```
Target dart2wasm failed: ProcessException: Process exited abnormally with exit code 64:
../../.pub-cache/hosted/pub.dev/permission_handler_html-0.1.3+4/lib/permission_handler_html.dart:3:1: Error: JS interop library 'dart:js_util' can't be imported when compiling to Wasm.
Try using 'dart:js_interop' or 'dart:js_interop_unsafe' instead.
import 'dart:js_util' as js_util;
```

### Steps to reproduce

1. Add permission_handler_html 0.1.3+4 as a dependency
2. Run flutter build web --wasm

### Code sample

...

### Current version

0.1.3+4

### Last version without regression

0.1.3+3

PR #1410, opened 2024-11-20
Fix for WASM compilation

dg76:
To fix https://github.com/Baseflow/flutter-permission-handler/issues/1409 I have changed the way how "mediaDevices" is loaded to make it compatible with WASM. Basically I am just using the same method now that was used for "permissions" a few lines below. It still seems to work fine for me regarding the bug https://github.com/Baseflow/flutter-permission-handler/issues/1386 .


dg76:
I think my branch was not correctly configured, it causes merge conflicts. I will try it again...

Issue #1455, opened 2025-03-08
[Enhancement proposal]: Update iOS minimum version to 10.0 or higher, delete dead fallback code


jmagman:
### Please check the following before submitting a new issue.

- [x] I have searched the [existing issues](https://github.com/baseflow/flutter-permission-handler/issues).
- [x] I have carefully [read the documentation](https://github.com/baseflow/flutter-permission-handler/blob/main/permission_handler/README.md) and verified I have added the required platform specific configuration.

### Please select affected platform(s)

- [ ] Android
- [x] iOS
- [ ] Windows

### Proposal

permission_handler_apple supports Flutter > 2.8.0, which had a minimum iOS requirement of iOS 9. 

https://github.com/Baseflow/flutter-permission-handler/blob/441c53c4e02ac9b6417f758495a4b04618c6e600/permission_handler_apple/pubspec.yaml#L9

This is a relevant code from the Flutter 2.8.0 tag that shows 9.0:
https://github.com/flutter/flutter/blob/cf4400006550b70f28e4b4af815151d1e74846c6/packages/flutter_tools/templates/app_shared/ios-objc.tmpl/Runner.xcodeproj/project.pbxproj.tmpl#L286
https://github.com/flutter/flutter/pull/62902

However, the permission_handler_apple podspec claims to support iOS 8 (both 8 and 9 are really old at this point):
https://github.com/Baseflow/flutter-permission-handler/blob/441c53c4e02ac9b6417f758495a4b04618c6e600/permission_handler_apple/ios/permission_handler_apple.podspec#L19

This isn't a problem in and of itself, but it means the iOS 8 fallback checks are always dead code.

Current Flutter supports iOS 12 and higher, so you could optionally bump further, but that would require increasing the minimum version of Flutter in the pubspec.

### Pitch

Increase the podspec to 10.0.

Remove the dead `if (@available(iOS 9.0, *))` and earlier check fallbacks

https://github.com/Baseflow/flutter-permission-handler/blob/441c53c4e02ac9b6417f758495a4b04618c6e600/permission_handler_apple/ios/Classes/PermissionManager.m#L89
https://github.com/Baseflow/flutter-permission-handler/blob/441c53c4e02ac9b6417f758495a4b04618c6e600/permission_handler_apple/ios/Classes/strategies/LocationPermissionStrategy.m#L177

jmagman:
[...]

PR #1461, opened 2025-03-28
Drop support for iOS < 11

stuartmorgan-g:
iOS versions < 11 haven't been supported since Flutter 3.3, which is over two years old. Even that version is too old to be able to ship to the App store with (as it lacks a privacy manifest), so there is no clear value in continuing to support versions even older than that.

This updates the Flutter SDK constraint to 3.3, where iOS 11 became the minimum, and removes all `@available` checks related to iOS 11 or earlier, as well as any fallback code that only ran on versions older than that.

Fixes https://github.com/Baseflow/flutter-permission-handler/issues/1455

## Pre-launch Checklist

- [x] I rebased onto `main`.

### DIFF
--- a/permission_handler_apple/ios/Classes/strategies/BluetoothPermissionStrategy.m
+++ b/permission_handler_apple/ios/Classes/strategies/BluetoothPermissionStrategy.m
@@ -61,10 +61,9 @@
 }
 
 - (void)handleCheckServiceStatusCallback:(CBCentralManager *)centralManager {
-    if (@available(iOS 10, *)) {
-        ServiceStatus serviceStatus = [centralManager state] == CBManagerStatePoweredOn ? ServiceStatusEnabled : ServiceStatusDisabled;
-        _serviceStatusHandler(serviceStatus);
-    }
+    ServiceStatus serviceStatus = [centralManager state] == CBManagerStatePoweredOn ? ServiceStatusEnabled : ServiceStatusDisabled;
+    _serviceStatusHandler(serviceStatus);
+
     #pragma clang diagnostic ignored "-Wdeprecated-declarations"
     ServiceStatus serviceStatus = [centralManager state] == CBCentralManagerStatePoweredOn ? ServiceStatusEnabled : ServiceStatusDisabled;
     _serviceStatusHandler(serviceStatus);
--- a/permission_handler_apple/ios/Classes/strategies/BluetoothPermissionStrategy.m
+++ b/permission_handler_apple/ios/Classes/strategies/BluetoothPermissionStrategy.m
@@ -63,10 +63,6 @@
 - (void)handleCheckServiceStatusCallback:(CBCentralManager *)centralManager {
     ServiceStatus serviceStatus = [centralManager state] == CBManagerStatePoweredOn ? ServiceStatusEnabled : ServiceStatusDisabled;
     _serviceStatusHandler(serviceStatus);
-
-    #pragma clang diagnostic ignored "-Wdeprecated-declarations"
-    serviceStatus = [centralManager state] == CBCentralManagerStatePoweredOn ? ServiceStatusEnabled : ServiceStatusDisabled;
-    _serviceStatusHandler(serviceStatus);
 }
 
 - (void)requestPermission:(PermissionGroup)permission completionHandler:(PermissionStatusHandler)completionHandler errorHandler:(PermissionErrorHandler)errorHandler {
--- a/permission_handler_apple/ios/Classes/strategies/BluetoothPermissionStrategy.m
+++ b/permission_handler_apple/ios/Classes/strategies/BluetoothPermissionStrategy.m
@@ -65,7 +65,7 @@
     _serviceStatusHandler(serviceStatus);
 
     #pragma clang diagnostic ignored "-Wdeprecated-declarations"
-    ServiceStatus serviceStatus = [centralManager state] == CBCentralManagerStatePoweredOn ? ServiceStatusEnabled : ServiceStatusDisabled;
+    serviceStatus = [centralManager state] == CBCentralManagerStatePoweredOn ? ServiceStatusEnabled : ServiceStatusDisabled;
     _serviceStatusHandler(serviceStatus);
 }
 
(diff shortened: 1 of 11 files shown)


permission_handler_apple/ios/Classes/strategies/BluetoothPermissionStrategy.m
```diff
@@ -61,12 +61,11 @@ - (void)checkServiceStatus:(PermissionGroup)permission completionHandler:(Servic
...
 
 - (void)handleCheckServiceStatusCallback:(CBCentralManager *)centralManager {
-    if (@available(iOS 10, *)) {
-        ServiceStatus serviceStatus = [centralManager state] == CBManagerStatePoweredOn ? ServiceStatusEnabled : ServiceStatusDisabled;
-        _serviceStatusHandler(serviceStatus);
-    }
+    ServiceStatus serviceStatus = [centralManager state] == CBManagerStatePoweredOn ? ServiceStatusEnabled : ServiceStatusDisabled;
+    _serviceStatusHandler(serviceStatus);
+
     #pragma clang diagnostic ignored "-Wdeprecated-declarations"
```
stuartmorgan, line 67: I left the code from here on to preserve behavior, but—with the caveat that I didn't look at the surrounding code—this looks very likely to be a pre-existing bug. It seems likely that this code was intended to be in an `else`, or after a `return` in the `if`, such that it only ran for iOS < 10. What it has actually been doing (and what this preserves) is running both sets of code on iOS 10+.

permission_handler_apple/ios/Classes/strategies/LocationPermissionStrategy.m
```diff
@@ -174,44 +174,36 @@ + (PermissionStatus)permissionStatus:(PermissionGroup)permission {
...
-                    return PermissionStatusRestricted;
-                case kCLAuthorizationStatusAuthorizedWhenInUse:
-                case kCLAuthorizationStatusDenied:
-                    return PermissionStatusPermanentlyDenied;
-                case kCLAuthorizationStatusAuthorizedAlways:
-                    return PermissionStatusGranted;
-            }
-        }
-        
+    if (permission == PermissionGroupLocationAlways) {
```
stuartmorgan, line 177: The diff for this method looks odd because of how the diff is computed. What I actually did is:
- removed the `if (@available(iOS 8.0, *)) {` wrapping, and decreased the indent level of the code inside
- trimmed the *last* `switch` to remove all the cases that were handled in the switch above.

The last two switches could be combined together, I just did the minimal change. I'm happy to restructure it if you would prefer.

permission_handler_apple/ios/Classes/strategies/SensorPermissionStrategy.m
```diff
@@ -13,12 +13,10 @@ - (PermissionStatus)checkPermissionStatus:(PermissionGroup)permission {
...
-          ? ServiceStatusEnabled
-          : ServiceStatusDisabled
-        );
-    }
+    completionHandler([CMMotionActivityManager isActivityAvailable]
+        ? ServiceStatusEnabled
+        : ServiceStatusDisabled
+    );
     
     completionHandler(ServiceStatusDisabled);
```
stuartmorgan, line 21: As with the first comment above, I left the existing behavior as-is, but I would be extremely surprised if this one is not a bug given that it's unconditionally reporting disabled right after the previous call reports the detected status. But as with the other, there was not an `else` or a `return`, so this was always running.

stuartmorgan (commented):
See notes below about potential bugs found during the cleanup, which this PR currently does not attempt to fix.

permission_handler_apple/ios/Classes/strategies/BluetoothPermissionStrategy.m
TimHoogstrate: Good found, It looks like `CBCentralManagerStatePoweredOn` is deprecated so we should remove this. I've tested this briefly and it still works nevertheless we should clean it up.

permission_handler_apple/ios/Classes/strategies/SensorPermissionStrategy.m
TimHoogstrate: Hi @stuartmorgan-g, I dove into this snippet and it indeed looks like a bug. HOWEVER, it is currently not possible to call this "code" because I noticed that the `sensor` permission is not an instance of `PermissionWithService`. So besides that we should remove the second completion handler, we should also update the sensor permission as a permissionWithService so that we can call this code. Maybe we should first merge these changes and later update the permission instance to a `PermissionWithService` in the platform_interface.

permission_handler_apple/ios/Classes/strategies/LocationPermissionStrategy.m
TimHoogstrate: Thanks for the update. It looks like `kCLAuthorizationStatusAuthorized` is deprecated for iOS but not for macOS (10.6+). Isn't it a better idea to get rid of the deprecation warning by wrapping the code in a compile-time check so that we know when it gets deprecated for macOS?

permission_handler_apple/ios/Classes/strategies/BluetoothPermissionStrategy.m
stuartmorgan-g: Okay, I've now removed the deprecated path.

permission_handler_apple/ios/Classes/strategies/LocationPermissionStrategy.m
stuartmorgan-g: That would change the behavior of this method, and would require fully auditing all codepaths that lead to it to determine whether the deprecated constant is ever passed in as a parameter on iOS, since if it is adding a compile-time check would cause it to return the wrong value on iOS.

Do you want to make that part of this PR, rather than changing behavior in a follow-up?

permission_handler_apple/ios/Classes/strategies/SensorPermissionStrategy.m
stuartmorgan-g: I've now made the minimal change of just removing this incorrect line.

permission_handler_apple/ios/Classes/strategies/LocationPermissionStrategy.m
TimHoogstrate: No let's keep this one simple. Let's do this in another PR.

TimHoogstrate (approved):
Nice work, thanks for your help!

PR #1518, opened 2026-02-13
[permission_handler_windows] Fix clang-cl compilation

Kataglyphis:
## Description

This PR fixes compilation of the Windows plugin with `clang-cl` (LLVM), which some projects use for faster Windows builds.

MSVC accepts the implicit conversion from `EncodableMap` to `EncodableValue` when calling `MethodResult::Success()`. `clang-cl` is stricter and rejects it.

### Reproduction

Configured the way a project on the LLVM toolset does — `/W3`, `-Wno-error` — compiling the plugin target against the released 0.2.2 package with no patches applied:

```
$ cmake --build . --target permission_handler_windows_plugin
permission_handler_windows_plugin.cpp(142,13): error: no matching member function for call to 'Success'
  142 |     result->Success(requestResults);
   29 |   void Success(const T& result) { SuccessInternal(&result); }
   33 |   void Success() { SuccessInternal(nullptr); }
1 warning and 1 error generated.
```

Applying only the `Success(flutter::EncodableValue(...))` change from this PR, and altering nothing else, that target compiles. This is an overload-resolution **error**, not a warning, so no `-Wno-...` flag makes it go away — a project on clang-cl currently cannot build this package without patching it locally.

## Changes

Scoped entirely to `permission_handler_windows`, and to that one file:

1. **Explicit type conversion** — wrapped `requestResults` in `flutter::EncodableValue(...)` before passing it to `result->Success()`.
   * Fixes: `no matching member function for call to 'Success'`
2. **Warning cleanup** for the `/W4 /WX` settings Flutter generates for every Windows app:
   * removed the unused `this` capture and the unused parameter names from the `Geolocator.PositionChanged` lambda — clang-cl reports these as `-Wunused-lambda-capture` and `-Wunused-parameter` (MSVC stays quiet on the latter because Flutter passes `/wd"4100"`, a quoted form clang-cl does not apply);
   * gave the request loop a `size_t` index, fixing the signed/unsigned comparison.
3. **Version** — bumped `permission_handler_windows` to `0.2.3` with a `CHANGELOG.md` entry.

The app-facing `permission_handler` package is deliberately **not** touched: its existing `permission_handler_windows: ^0.2.2` constraint already allows 0.2.3.

## Testing

* [x] Compiles with **clang-cl** — verified in a project that builds its whole Flutter app with the LLVM toolset (https://github.com/Kataglyphis/Kataglyphis-Inference-Engine): the plugin compiles from these sources with no local patching, and its DLL is installed into the app bundle. That project had been carrying a local patch for exactly this `Success()` call, which these changes make unnecessary.
* [x] The changed code is **MSVC**-clean too: with the `/await` line removed locally, `flutter build windows --release` of the example builds the whole app on MSVC 14.51. Left in place, the package still fails there for the pre-existing reason in note 2 below — that is not something this PR changes either way.
* [x] Verified `requestPermissions` still returns correct results on Windows.

> **Two unrelated problems I ran into while verifying this, both already on `main` and neither addressed here** — happy to open separate PRs if useful:
> 1. `permission_handler_windows/example` cannot build: its `pubspec.yaml` still declares `packages/baseflow_plugin_template/logo.png` and `poweredByBaseflow.png`, but `baseflow_plugin_template` 2.2.0 moved those out of `lib/` and declares them itself, so `flutter build windows` stops at `No file or variants found for asset` before the compiler runs. The `apple`, `android` and `html` examples already dropped those two lines.
> 2. `windows/CMakeLists.txt` passes `/await`. On MSVC 14.51 that now fails the build outright — `error C2338: static assertion failed: 'error STL1011: The /await compiler option ... will be REMOVED SOON'` — and under Flutter's default `/W4 /WX` clang-cl rejects it as an unused argument. The target already requires C++20, so `<coroutine>` supersedes it and the flag can simply go.

> **Note**
> This branch was rebuilt on top of current `main` to resolve an earlier merge conflict. A previous revision of this PR bumped `permission_handler` to `12.0.2` and `permission_handler_windows` to `0.2.2`; both have since been released upstream with different contents, so those version changes were dropped.

## Related Issue

Fixes build failures for users compiling the Windows plugin with `clang-cl`.

## Pre-launch Checklist

- [x] I updated `pubspec.yaml` with an appropriate new version according to the [pub versioning philosophy].
- [x] I rebased onto `main`.

### DIFF
--- a/permission_handler/CHANGELOG.md
+++ b/permission_handler/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 12.0.2
+
+- Updated windows permission handler to be compatible with clang-cl
+
 ## 12.0.1
 
 - Updates the correspondence between permission groups and the key values of Info.plist in the README.md.
--- a/permission_handler/pubspec.yaml
+++ b/permission_handler/pubspec.yaml
@@ -27,7 +27,7 @@ dependencies:
   permission_handler_android: ^13.0.0
   permission_handler_apple: ^9.4.6
   permission_handler_html: ^0.1.1
-  permission_handler_windows: ^0.2.1
+  permission_handler_windows: ^0.2.2
   permission_handler_platform_interface: ^4.3.0
 
 dev_dependencies:
--- a/permission_handler/pubspec.yaml
+++ b/permission_handler/pubspec.yaml
@@ -2,7 +2,7 @@ name: permission_handler
 description: Permission plugin for Flutter. This plugin provides a cross-platform (iOS, Android) API to request and check permissions.
 repository: https://github.com/baseflow/flutter-permission-handler
 issue_tracker: https://github.com/Baseflow/flutter-permission-handler/issues
-version: 12.0.1
+version: 12.0.2
 
 environment:
   sdk: ^3.5.0
--- a/permission_handler_windows/CHANGELOG.md
+++ b/permission_handler_windows/CHANGELOG.md
@@ -1,3 +1,8 @@
+## 0.2.2
+
+* clang-cl support
+* Explicit Type Conversion: Wrapped `requestResults` in `flutter::EncodableValue(...)` before passing it to `result->Success()
+
 ## 0.2.1
 
 * Updates the dependency on `permission_handler_platform_interface` to version 4.1.0 (SiriKit support is only available for iOS and macOS).
--- a/permission_handler_windows/CHANGELOG.md
+++ b/permission_handler_windows/CHANGELOG.md
@@ -1,3 +1,8 @@
+## 0.2.3
+
+* Fixes compilation with `clang-cl` (LLVM) by explicitly wrapping the request results map in a `flutter::EncodableValue` before passing it to `result->Success()`.
+* Removes an unused `this` capture from the `Geolocator.PositionChanged` lambda, which triggered `-Wunused-lambda-capture`.
+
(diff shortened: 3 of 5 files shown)


Kataglyphis:
Rebuilt this branch on top of current `main` to resolve the merge conflict, and squashed it into a single commit.

Summary of what changed since the last revision:

- **Dropped the version bumps that caused the conflict.** `permission_handler_windows` `0.2.2` and `permission_handler` `12.0.2`/`13.x` have all been released upstream since this PR was opened, so the original bumps collided with released versions. The Windows package is now bumped to `0.2.3` instead.
- **Removed all changes to the app-facing `permission_handler` package.** The existing `permission_handler_windows: ^0.2.2` constraint already permits `0.2.3`, so no bump is required there. This was also the cause of the failing `App facing package` check: the PR pointed the constraint at `^0.2.2` while `0.2.2` was not yet on pub.dev, so `flutter pub get` could not resolve. With `permission_handler/**` untouched, that path-filtered workflow no longer runs for this PR.
- @Copilot's note about the unterminated backtick in the changelog is addressed — the entry was rewritten.

The PR is now three files, all inside `permission_handler_windows`: the two-line C++ fix, the version bump, and the changelog entry. No conflicts remain.

PR #1547, opened 2026-07-13
Migrate permission_handler_apple to Pigeon ProxyApi (darwin)

rickdijk:
- Migrate iOS from legacy Obj-C MethodChannel strategies to Pigeon ProxyApi
- Add `darwin/` Swift plugin (thin bridges) and `lib/next/` Dart permission logic
- Enable cutover via `sharedDarwinSource: true` and `dartPluginClass: PermissionHandlerApple`
- Bump `permission_handler_apple` to 9.5.0 (minimum iOS 14.0 for darwin)

## Pre-launch Checklist

- [x] I rebased onto `main`.


permission_handler_apple/example/pubspec.yaml
```diff
@@ -9,6 +9,7 @@ dependencies:
   flutter:
     sdk: flutter
 
+  permission_handler_platform_interface: any
```
mvanbeusekom, line 12: Is this dependency really necessary? I don't really see why the example app should depend on the permission_handler_platform_interface.

permission_handler_apple/lib/next/src/permission_manager.dart
mvanbeusekom, line 1: I think we should simplify the design and merge the `PermissionManager` and the `PermissionHandlerAppleImpl` into the `PermissionHandlerApple` class.

The Obj-C `PermissionManager` class was originally created to keep track of state (are we already requesting permissions). This was mainly done because we could only hold one reference of the `FlutterResult` instance (which maps the instance of the `MethodChannel` on the Dart side with the instance on the native side). With the new `ProxyApi` and the special `InstanceManager` it creates this is no longer an issue and can keep track of multiple instances (read permission requests) at the same time. So we no longer need to keep track of the fact if we are already requesting permissions or not and can safely remove all the logic around the `_requestInProgress` boolean making this class a lot simpler and therefore easy to merge with the `PermissionHandlerAppleImpl` and `PermissionHandlerApple` classes.

mvanbeusekom (changes requested):

PR #1553, opened 2026-08-07
Fix SPM Info.plist discovery for flavored and multi-configuration apps (1/2)

### LINKED ISSUE #1548: [Bug] Automatic permission detection doesn't work in apps using flavors
Versions: Flutter 3.44.4, permission_handler_apple 9.4.10

The automatic detection of permissions added with the support for Swift Package Manager doesn't work for applications using multiple Info.plist files separated by build configurations (e.g., Info-Debug.plist and Info-Release.plist) instead of the default Info.plist.

It seems the plugin fails to detect the required permission keys because it only searches for the hardcoded ios/Runner/Info.plist path during the build time.

NOTE: originally reported by @tazan83 as part of issue #1544.

Pachebel:
> First of two PRs. #1554 (per-flavor permissions) is stacked on this branch and
> should be merged after it.

`Package.swift` derives the `PERMISSION_*` macros from the host app's `Info.plist`.
Discovery failed in two independent ways, and the failure mode is silent: every
permission is compiled out and permission checks report `denied` with no
indication of why.

**1. Hardcoded path.** The manifest only ever looked at `ios/Runner/Info.plist`.
Apps using build-configuration or flavor specific plists (`Info-Debug.plist`,
`Runner/Info-$(CONFIGURATION).plist`, …) were never matched — the layout reported
in #1548.

**2. Broken walk-up.** The app-root lookup required a `pubspec.yaml` and a
loadable plist in the same condition, so a miss did not stop the walk-up at the
app and could adopt an unrelated app's plist in a monorepo.

## What this does

- Resolves `Info.plist` locations from `INFOPLIST_FILE` in the Xcode project and
  any `.xcconfig` files, with a scan of `ios/` as a fallback, and merges the
  usage description keys found across them. A divergence warning fires when
  configurations declare different keys, since the merged union ends up in every
  binary (ITMS-90683).
- Anchors the app root on a `pubspec.yaml` next to an `ios/*.xcodeproj`, which is
  what distinguishes a host app from a plugin package.
- Hardens the `INFOPLIST_FILE` parsing: the pattern cannot match inside
  `GENERATE_INFOPLIST_FILE` (which Xcode writes into every templated target), and
  candidates that are not files on disk do not suppress the fallback scan.
  Either one alone was enough to compile out every permission for an app whose
  target names its plist through an unexpanded build variable.
- Adds `PERMISSION_HANDLER_INFO_PLIST` to point the manifest at specific plists —
  the only mechanism available to builds started from Xcode.app, which run with
  `/` as their working directory and receive none of Xcode's build settings.
- Adds `PERMISSION_HANDLER_VERBOSE` to log the app root, the plists used and the
  resolved macros, because Xcode discards manifest output entirely.

## Testing

The example's Debug configuration now uses a separate `Info-Debug.plist`,
identical to `Info.plist` except that it declares no
`NSContactsUsageDescription`. `PERMISSION_CONTACTS=1` is therefore only reachable
by discovering and merging both plists, which is what lets CI assert on the
discovery itself rather than on the example's default plist. A second CI step
builds a synthetic project that forces the scan fallback.

No runtime code is touched: the compiled-out strategies already degrade to
`denied` through `UnknownPermissionStrategy`, and CocoaPods builds are unaffected.

Fixes #1548

## Pre-launch Checklist

- [x] I rebased onto `main`.


Pachebel:
Let me know if there are any questions!

Thanks for letting me contribute. :)

PS: I put an 1/2 in the PR title, because I'm going to make the second one as we talk before

mvanbeusekom (approved):
LGTM! 

Thanks for this PR, changes look really good. I really appreciate the debugging feature using the `PERMISSION_HANDLER_VERBOSE` to validate the different paths and detect the differences between the different configurations when using flavors.

Issue #1544, opened 2026-06-13
Swift Package manager iisue


lahteefmitco:
Noit supporting Swift Package manager in flutter version 3.44. please solve it

jendas1:
Facing the same issue which creates problem with the flutter workflow. Now CocoaPods are deprecated by flutter (as mentioned here: https://docs.flutter.dev/add-to-app/ios/project-setup ).

daniJimen:
The problem is that SPM support was updated in permission_handler_apple 9.4.10. However, permission_handler 12.0.3 still does not reference the latest version of permission_handler_apple, so those SPM updates are not included in the current release of permission_handler.

lyio:
@daniJimen is that a simple version bump or does the update require changes?

Pachebel:
Guys, I have fixed this "issue" removing any permissions I have made previously on Podfile. Here is some sort of step-by-step guide:

## 1. Upgrade Flutter Packages
To ensure you have the SPM-compatible version of the Apple permission handler, you need to upgrade your packages.

```bash
flutter pub upgrade
```
*(Ensure `permission_handler_apple` updates to version `9.4.10` or higher).*

## 2. Remove Old CocoaPods Configurations from `Podfile`
With SPM, `Package.swift` automatically detects which permissions to enable by reading the keys present in your app's `ios/Runner/Info.plist`. You no longer need to manually inject C macros in the `Podfile`.

Open `ios/Podfile` and **remove** the entire `GCC_PREPROCESSOR_DEFINITIONS` block associated with permissions.

**Remove this block (if present):**
```ruby
# Remove these lines from your post_install loop
config.build_settings['GCC_PREPROCESSOR_DEFINITIONS'] ||= [
  '$(inherited)',
  'PERMISSION_CAMERA=1',
  'PERMISSION_PHOTOS=1',
  'PERMISSION_LOCATION=1',
  'PERMISSION_NOTIFICATIONS=0',
  # ... and any other PERMISSION_ macros
]
```

## 3. Opt-out of Default Permissions (Notifications / Critical Alerts)
Since `PermissionGroup.notification` has no required `Info.plist` key, it is **enabled by default**. If your app does not use notifications (previously `PERMISSION_NOTIFICATIONS=0`), you now must disable it via an environment variable before building.

In the terminal:
```bash
export PERMISSION_NOTIFICATIONS=0
flutter build ios
```

## 4. Clear Xcode Derived Data
[...]

PR #1554, opened 2026-08-07
Add opt-in per-flavor permissions for Swift Package Manager builds (2/2)

### LINKED ISSUE #1548: [Bug] Automatic permission detection doesn't work in apps using flavors

Pachebel:
> **Stacked on #1553 — do not merge before it.**
> This branch is built on top of `fix/spm-info-plist-flavors`, so until #1553 lands the diff shown
> here contains both commits. Only the second one, `Add opt-in per-flavor permissions for Swift
> Package Manager builds`, belongs to this PR. I will rebase onto `main` and force-push once #1553
> merges, after which the diff will show this change alone.
>
> This is the second of the two PRs discussed in #1548 — the first fixes plist discovery, this one
> adds the per-flavor opt-in.

### The problem

#1553 makes the manifest find every `Info.plist` an app has, and merges their usage description
keys. Merging is the only thing a manifest can do on its own, and for apps with per-flavor plists it
enables too much: if `Info-dev.plist` declares `NSCameraUsageDescription` and `Info-prod.plist` does
not, the camera code is still compiled into the production binary. That is exactly the
`ITMS-90683` rejection these macros exist to prevent, so for flavored apps the merge turns a build
convenience into an App Store risk.

### The change

An app can declare a `permission_handler.json` next to its `pubspec.yaml`:

```json
{
  "strict": true,
  "flavors": {
    "dev":  { "infoPlist": "ios/Runner/Info-dev.plist",
              "configurations": ["Debug-dev", "Profile-dev", "Release-dev"] },
    "prod": { "infoPlist": "ios/Runner/Info-prod.plist",
              "configurations": ["Debug-prod", "Profile-prod", "Release-prod"] }
  }
}
```

Each flavor names the single `Info.plist` that defines it. The manifest reads only the selected
flavor's plist and merges nothing, so a flavor cannot inherit another flavor's permissions — that is
a property of the construction, not a check that could be forgotten. Apps without the file keep the
behaviour from #1553 unchanged.

A flavor declares an `infoPlist` rather than a list of permission names on purpose. A list like
`["camera", "microphone"]` would require the permission vocabulary to be duplicated in the manifest,
in the `select` command and in the verification script, and any drift between those three copies
would fail silently. The plist stays the single source of truth, which is also what Apple reads.

With `"strict": true` (the default) a build whose flavor cannot be determined compiles in no
permissions at all, rather than falling back to the union.

A flavor whose `infoPlist` path does not exist is an error that compiles no permissions in, rather
than a fallback to discovery — falling back would hand that flavor the union of every other
flavor's permissions, which is the leak this config exists to prevent.

### Why selecting a flavor is a separate step

The obvious question is why the manifest does not just detect the active flavor. I instrumented it
and measured three things rather than assuming them:

1. **Xcode passes none of its build settings to manifest evaluation.** No `CONFIGURATION`, `SRCROOT`,
   `PROJECT_DIR` or `INFOPLIST_FILE`. The manifest process' argv carries only the canonical
   `packageDirectory`. So the manifest cannot know which configuration, and therefore which flavor,
   is being built.
2. **Xcode does not re-evaluate a cached manifest when an environment variable changes.** Building
   with `PERMISSION_CONTACTS=0` and then with `PERMISSION_CONTACTS=1` against warm DerivedData
   produces the *first* result both times, and the same in reverse. The value only changes once
   `SourcePackages`, `Build/Intermediates.noindex/XCBuildData` and the shared SwiftPM manifest cache
   are removed — clearing `SourcePackages` alone is not enough. (The `swift package` CLI does
   re-evaluate, which is why this is easy to miss.)
3. **Scheme pre-actions run after package resolution.** A pre-action that writes a file the manifest
   reads has no effect on that same build, so the selection cannot be prepared automatically from
   inside Xcode.

Together those mean the flavor has to be chosen before the build, and the caches have to be cleared
when it changes. That is what the new command does:

```bash
dart run permission_handler_apple:select prod   # --list shows what is declared
flutter run --flavor prod
```

It validates the config, records the choice in `ios/Flutter/permission_handler.selected`, clears
exactly the three cache locations above, and prints the usage descriptions that will be compiled in.

### Why there is also a build phase

Point 2 has a consequence worth stating plainly: **a cached manifest is never executed**, so no
check inside `Package.swift` can notice that its own result went stale. `strict` catches a missing
selection on the builds where the manifest actually runs, but it cannot catch a selection that
changed after the fact.

The only code that runs on every build *and* can see `CONFIGURATION` is a build phase on the app
target. `tool/verify_flavor_selection.sh` compares the recorded flavor against the configuration
being built and fails with an actionable message:

```
error: [permission_handler_apple] building "Release-prod" needs the "prod" permission flavor,
but "dev" is selected, so this build would ship dev's permissions.
Run: dart run permission_handler_apple:select prod
```

It is a no-op for projects without a `permission_handler.json` and for CocoaPods builds, where the
macros come from the `Podfile` and a selection means nothing. It is added by the user — a Swift
package cannot inject a build phase into the host project the way a CocoaPods `post_install` hook
can — and belongs at the top of the phase list, so a mismatch fails in a second instead of after a
full compile.

### Verification

The example app now declares two flavors over two `Info.plist` files that differ by exactly one key:
`debug` (`Info-Debug.plist`) has no contacts entry, `release` (`Info.plist`) does. That single
difference is what makes isolation observable.

- Manifest resolution: `debug` → `PERMISSION_CONTACTS=0`, `release` → `PERMISSION_CONTACTS=1`, with
  camera and microphone enabled in both.
- `strict` with no selection, and an undeclared flavor name, both resolve every macro to `0` and
  emit an error.
- End to end through `select` and `xcodebuild`, with no environment variables set, alternating
  flavors against the same DerivedData: contacts symbols go absent → present → absent.
- The build phase: a `Debug` build with `release` selected fails with exit 65 and the message above;
  an aligned build succeeds.
- A real `flutter build ios --debug` on the `debug` flavor produces a binary with no
  `CNContactStore` references but with `AVCaptureDevice` — the permission excluded by the flavor is
  genuinely absent from the shipped code, not merely reported as denied.

CI asserts the three resolution cases, that the pre-flavor merging path from #1553 still resolves
and merges both plists when no config applies, and that a mismatched selection fails the build.

Fixes #1548, #1544

### Notes for review

- **This extends the CI step added in #1553 rather than replacing it.** The merge assertions from
  that PR still run, with the config taken out of the picture via `PERMISSION_HANDLER_CONFIG`, and
  the per-flavor assertions run alongside them — so a regression in either path fails the build.
- **The repository `.gitignore` excluded every `bin/` directory**, which would have silently dropped
  `bin/select.dart` from the published package and broken
  `dart run permission_handler_apple:select` for everyone. This adds a narrow exception; pub
  requires package executables to live in `bin/`.
- Version bumped to `9.6.0` — new feature, no breaking change. Apps without a
  `permission_handler.json` are unaffected.
- Switching flavors costs a partial rebuild, since the manifest and the package targets have to be
  re-resolved. That is inherent to how SwiftPM caches manifests, not something this design chose.

## Pre-launch Checklist

- [ ] I rebased onto `main`. — deliberately not yet: this is stacked on #1553. Rebasing now would
      duplicate its commits. I will rebase and force-push as soon as #1553 merges.


Pachebel:
⚠️ Depends on https://github.com/Baseflow/flutter-permission-handler/pull/1553 ⚠️

mvanbeusekom (commented):
The changes and concept looks really good, I think this would be a really great addition to the plugin. 

One question though, would it be possible to convert the `permission_handler.json` configuration file to YAML format. Personally I'd prefer the more explicit JSON configuration, however since Flutter itself heavily relies on YAML configuration (pubspec.yaml), I feel it will feel more natural to Flutter developers if this configuration also done through YAML.

Pachebel:
> The changes and concept looks really good, I think this would be a really great addition to the plugin.
> 
> One question though, would it be possible to convert the `permission_handler.json` configuration file to YAML format. Personally I'd prefer the more explicit JSON configuration, however since Flutter itself heavily relies on YAML configuration (pubspec.yaml), I feel it will feel more natural to Flutter developers if this configuration also done through YAML.

I totally agree about the YAML format! 

One implementation note so the shape of the change makes sense: a Swift package manifest can't parse YAML natively. Foundation has no YAML support, and `Package.swift` can't import any library for its own evaluation. The same goes for the verification build phase (`yaml` isn't in the Python standard library). So instead of hand-rolling YAML parsers in three places, I'd like to structure it like this:

- The developer only ever touches `permission_handler.yaml`, right next to `pubspec.yaml`, exactly as you suggested.
- `dart run permission_handler_apple:select` reads it with `package:yaml` (the same parser pub itself uses) and generates `ios/Flutter/permission_handler.resolved.json`. It will be gitignored, never edited by hand just as Flutter's `Generated.xcconfig`.
- The manifest and the build phase keep reading that generated JSON with the native parsers they already have (`JSONSerialization` / Python's `json`).

Since `select` is already a required step before any flavored build (because of the manifest caching described in the PR), this adds no new workflow, the JSON just becomes an internal artifact instead of the user-facing format.

I think that we will need two guards against the file going stale: 
1. The build phase fails when the YAML is newer than the generated JSON ("re-run select")
2. A config without a generated file compiles no permissions in and says why. So editing the YAML and forgetting to re-run `select` can't silently ship the wrong permissions (think about all the devs that we will save with that 😆).

How does this sound to you? (I will implement it as it is low effort, and we can merge it or change if you have a better approach).

Again, thanks for the suggestion!

mvanbeusekom (approved):
LGTM! Thank you for going out of your way and make YAML configuration available. I was already afraid Swift / Python would not support it out of the box and don't allow external packages during the build phase. I really like the approach of parsing the YAML file using the `permission_handler_apple:select` step and converting it to a temporary JSON file which can be used by Swift during the rest of the build phase.

All in all the code looks really good and I think this will be a great solution, helping out a lot of users that are currently stuck simply because they use multiple flavors in their build pipeline.

I have one more request, would you mind having a look at the "iOS - Swift Package Manager" section in the [permission_handler/README.md](https://github.com/Baseflow/flutter-permission-handler/blob/main/permission_handler/README.md) file and see if it still covers the necessary details? Possibly we can reference to the README.md in the permission_handler_apple package to guide users to more detailed information? Since the permission_handler_apple package is not listed, not a lot of users will easily find the great documentation you have already written there. These changes could be part of a separate PR. In the meantime I will go ahead and merge and publish this new version.

Pachebel:
Of course I can, @mvanbeusekom.
Thanks for the compliments and review!

Doc changes: https://github.com/Baseflow/flutter-permission-handler/pull/1555

PR #1560, opened 2026-09-04
docs: clarify which Info.plist key to delete #1013

### LINKED ISSUE #1013: Doc clarification: Delete the (which?) *corresponding* permission description in Info.plist?
On

https://pub.dev/packages/permission_handler

Setup: iOS

From point 2, if I have

```
## dart: PermissionGroup.calendar
        'PERMISSION_EVENTS=1',

```
in Podfile,

Then in point 3, I need to delete the *corresponding* (???) NSCalendarsUsageDescription from Info.plist?

Is this conflict with the first sentence:

```
Add permission to your Info.plist file. Here's an example Info.plist with a complete list of all possible permissions.
```

(BTW, for iOS, the point 3 the camera permission example is confusing, I think it should continue just using the calendar permission example from point 2).

AzazelSensei:
The CocoaPods iOS setup said to delete the "corresponding" Info.plist key after a calendar Podfile example, then used camera as the plist example. I kept calendar for both steps and spelled out that the key to remove is the one for the permission you set to 0.

Fixes #1013


mvanbeusekom (approved):

mvanbeusekom:
Looks good, thanks for taking the time to clear this up.

PR #1566, opened 2026-09-12
fix(apple): prevent double completionHandler call in PhonePermissionStrategy

### LINKED ISSUE #1565: [Bug]: PhonePermissionStrategy.checkServiceStatus calls completionHandler twice, causing crashes
### Please check the following before submitting a new issue.

- [x] I have searched the [existing issues](https://github.com/baseflow/flutter-permission-handler/issues).
- [x] I have carefully [read the documentation](https://github.com/Baseflow/flutter-permission-handler/blob/main/permission_handler/README.md) and verified I have added the required platform specific configuration.

### Please select affected platform(s)

- [ ] Android
- [x] iOS
- [ ] Windows

### Steps to reproduce

1. Call `Permission.phone.serviceStatus` (or anything that triggers `checkServiceStatus` for the phone permission) on a device/simulator where `[[UIApplication sharedApplication] canOpenURL:[NSURL URLWithString:@"tel://"]]` returns `NO` (e.g. iPad, or a device without telephony capability).
2. In `PhonePermissionStrategy.m`, the `if (![app canOpenURL:telURL])` branch calls `completionHandler(ServiceStatusNotApplicable)` but does not `return`, so execution falls through and calls `completionHandler(...)` a second time with the enabled/disabled status.
3. The Flutter method channel result callback backing `completionHandler` is only expected to be invoked once; the second invocation raises an exception/crash.

We noticed this because `checkServiceStatus`/`PhonePermissionStrategy` shows up as one of the **top crashes in our production consumer app in Firebase Crashlytics**, affecting users on devices where `canOpenURL:` returns `NO` for `tel://`.

### Expected results

`completionHandler` should be invoked exactly once per call to `checkServiceStatus`.

### Actual results

`completionHandler` is invoked twice when the device cannot open `tel://` URLs, which can crash the app.

### Code sample

<details><summary>Code sample</summary>

```objc
// permission_handler_apple/ios/permission_handler_apple/Sources/permission_handler_apple/strategies/PhonePermissionStrategy.m

- (void)checkServiceStatus:(PermissionGroup)permission completionHandler:(ServiceStatusHandler)completionHandler {
  UIApplication *app = [UIApplication sharedApplication];
  NSURL *telURL = [NSURL URLWithString:@"tel://"];
  if (![app canOpenURL:telURL]) {
      completionHandler(ServiceStatusNotApplicable);
      // missing `return;` here — falls through and calls completionHandler again below
  }
  completionHandler([self canDevicePlaceAPhoneCall] ? ServiceStatusEnabled : ServiceStatusDisabled);
}
```

</details>

### Screenshots or video

<details>
<summary>Screenshots or video demonstration</summary>

Crashlytics shows this as a top crash in our production app; no visual repro available since it's a native-side double-callback crash.

</details>

### Version

9.6.1 (permission_handler_apple)

### Flutter Doctor output

<details><summary>Doctor output</summary>

```console
N/A — issue is isolated to the native iOS implementation and reproducible by code inspection, independent of Flutter/toolchain version.
```

</details>

ryanaidilp:
## Summary
- Fixes #1565
- `checkServiceStatus` in `PhonePermissionStrategy.m` called `completionHandler(ServiceStatusNotApplicable)` inside the `if (![app canOpenURL:telURL])` branch without returning, so execution fell through and called `completionHandler` a second time with the enabled/disabled status.
- This double invocation of the result callback shows up as a top crash in production (via Firebase Crashlytics) on devices where `canOpenURL:` returns `NO` for `tel://` (e.g. iPad).
- Adds the missing `return;` so `completionHandler` is only ever invoked once.

## Test plan
- [x] Read through every other `*PermissionStrategy.m` file under `permission_handler_apple` to confirm no other strategy has the same missing-`return` fallthrough pattern (all others either early-`return` after each handler call, or have a single-statement stub body).
- [x] Manual verification on a device/simulator where `canOpenURL:` returns `NO` for `tel://` (e.g. iPad) to confirm `checkServiceStatus` for `Permission.phone` no longer crashes.


ryanaidilp:
@mvanbeusekom can you please review this PR? Thanks

ibrahim-iqbal (commented):
Confirmed the bug on `master` — `PhonePermissionStrategy.m:22-24` invokes `completionHandler(ServiceStatusNotApplicable)` inside the `if (![app canOpenURL:telURL])` branch and then falls straight through to the trailing `completionHandler([self canDevicePlaceAPhoneCall] ? ServiceStatusEnabled : ServiceStatusDisabled)` on line 25. Two invocations of the same handler for the branch where `tel://` is not openable, which is the iPad case in #1565. Adding `return;` after the first call is the minimal fix and stops the extra `canDevicePlaceAPhoneCall` CoreTelephony work at the same time.

Cross-checked the neighbouring strategies to back the "no other strategy has the same fall-through" claim from the PR body:

- `AppTrackingTransparencyPermissionStrategy.m` — every `if` branch that calls the handler `return`s.
- `AssistantPermissionStrategy.m` — single-statement body.
- `AudioVideoPermissionStrategy.m`, `BackgroundRefreshStrategy.m`, `BluetoothPermissionStrategy.m` — same shape, each handler-invoking branch is followed by a `return` or is the terminal statement.

So this is genuinely the only offender.

One optional readability nudge, take or leave: `if (![app canOpenURL:telURL]) { completionHandler(ServiceStatusNotApplicable); return; } else { completionHandler(...); }` or an `else { ... }` around the trailing call would make the "these two are mutually exclusive" contract structural rather than implied by control flow. `return;` alone works and is the smallest possible diff, which is probably the right trade for a hot-fix on a production crash.

ryanaidilp:
> Confirmed the bug on `master` — `PhonePermissionStrategy.m:22-24` invokes `completionHandler(ServiceStatusNotApplicable)` inside the `if (![app canOpenURL:telURL])` branch and then falls straight through to the trailing `completionHandler([self canDevicePlaceAPhoneCall] ? ServiceStatusEnabled : ServiceStatusDisabled)` on line 25. Two invocations of the same handler for the branch where `tel://` is not openable, which is the iPad case in #1565. Adding `return;` after the first call is the minimal fix and stops the extra `canDevicePlaceAPhoneCall` CoreTelephony work at the same time.
> 
> Cross-checked the neighbouring strategies to back the "no other strategy has the same fall-through" claim from the PR body:
> 
> * `AppTrackingTransparencyPermissionStrategy.m` — every `if` branch that calls the handler `return`s.
> * `AssistantPermissionStrategy.m` — single-statement body.
> * `AudioVideoPermissionStrategy.m`, `BackgroundRefreshStrategy.m`, `BluetoothPermissionStrategy.m` — same shape, each handler-invoking branch is followed by a `return` or is the terminal statement.
> 
> So this is genuinely the only offender.
> 
> One optional readability nudge, take or leave: `if (![app canOpenURL:telURL]) { completionHandler(ServiceStatusNotApplicable); return; } else { completionHandler(...); }` or an `else { ... }` around the trailing call would make the "these two are mutually exclusive" contract structural rather than implied by control flow. `return;` alone works and is the smallest possible diff, which is probably the right trade for a hot-fix on a production crash.

Good point on readability. I went this route to keep the code changes minimal, but I agree your approach is better. I'll wrap it in an else block and push the update. Thanks

